PR #11, opened 2026-03-08
Detect JSON key renames when value is structurally identical

### LINKED ISSUE #10: JSON differ should detect key renames when value is structurally identical
When a JSON key is renamed but its value stays the same, the differ reports it as a deletion and an addition rather than a rename.

Example:                                                                                                                                            
// before
`{ "timeout": 30 }
`
// after
`{ "request_timeout": 30 }
`
Current output:
```
property   timeout          [deleted]
property   request_timeout  [added]
```

Expected:
` property   request_timeout  [renamed from timeout]
`

This should work for scalar values but also nested objects. If an object is moved under a new key with identical contents, the same problem occurs.

The function `differ` already handles this by hashing the body and matching pairs with identical hashes across the deleted and added sets. The same approach could work here as a post-processing pass on the JSON diff results, and `xxhash` is already a dependency.

nminev:
Fixes #10.

Currently the JSON differ always reports a renamed key as a deletion + addition, even when the value is unchanged. This change fixes that by computing a `structural_hash` over just the value portion of each JSON entity (stripping the key name). The existing Phase 2 hash-based matching in `match_entities()` already uses `structural_hash` as a fallback for rename detection, so no changes to the matching logic were needed.

Before:
```
⊖ property   timeout          [deleted]
⊕ property   request_timeout  [added]
```

After:
```
↻ property   request_timeout  [renamed]
```

Works for both scalar values and nested objects.

Two new tests added to `json.rs` covering the scalar and object cases.

### DIFF
--- a/crates/sem-core/src/parser/plugins/json.rs
+++ b/crates/sem-core/src/parser/plugins/json.rs
@@ -239,6 +239,8 @@ fn trim_trailing_blanks(lines: &[&str], start: usize, next_start: usize) -> usiz
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::model::change::ChangeType;
+    use crate::model::identity::match_entities;
 
     #[test]
     fn test_json_line_positions() {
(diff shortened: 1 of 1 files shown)


rs545837 (approved):
Clean approach, thanks for the PR! Leveraging structural_hash for JSON rename detection is the right call.

PR #14, opened 2026-03-08
fix sem diff replaying the last commit after commit

MattCozendey:
When sem diff found no staged or working tree changes, it fell back to HEAD and returned the last commit diff. That made a clean repo look like it still had pending changes and read like stale cached output.

Repro:
1. Modify a tracked file
2. Run sem diff
3. Commit the change
4. Run sem diff again

Expected: no changes detected
Actual: the same semantic diff is shown again from HEAD

Stop the clean-worktree path from auto-switching to commit scope. Explicit history queries still go through --commit, --from, and --to.

Also add a regression test covering the clean-worktree case and preserving explicit commit diff behavior.

### DIFF
--- a/crates/sem-core/src/git/bridge.rs
+++ b/crates/sem-core/src/git/bridge.rs
@@ -71,16 +71,8 @@ impl GitBridge {
             return Ok((DiffScope::Working, working_files));
         }
 
-        // Fall back to HEAD commit
-        match self.get_head_sha() {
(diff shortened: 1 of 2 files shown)


rs545837 (approved):
Clean fix. The fallback to HEAD on a clean worktree was definitely wrong. Tests cover both the fix and the explicit commit path. Thanks for catching this.

PR #21, opened 2026-03-09
feat: Refine JS/TS binding entity classification for const/function/generator

MattCozendey:
JS/TS binding declarations were not being classified with the right level of specificity.

Before this change:
- plain `const` bindings in JS/TS were flattened to `variable`
- function-valued bindings were only partially refined, and the behavior was not consistent across the TypeScript parser and the Rust core
- generator-valued bindings were not surfaced as a distinct entity type
- review risk detection only treated `function` and `method` deletions as function-like removals

After this change:
- `const foo = "abc"` is classified as `constant`
- `let foo = "abc"` and `var foo = "abc"` stay `variable`
- `const foo = function () {}` is classified as `function`
- `const foo = () => {}` is classified as `function`
- `let foo = function* () {}` is classified as `generator`
- function-like classification wins over declaration kind, so `const` does not mask a more specific function/generator entity type
- deleted `generator` entities are included in review risk signals the same way deleted functions already are

Implementation notes:
- added a small declaration classifier in the TS parser that first checks whether a binding's assigned value is function-like and only falls back to declaration kind (`const` vs `let`/`var`) when it is not
- mirrored the same logic in `sem-core` so the Rust path and TS path stay behaviorally aligned
- kept local non-function bindings inside function bodies suppressed, so this does not widen entity extraction noise
- added focused parser tests for both `.ts` and `.js` inputs covering constant, variable, function expression, arrow function, and generator cases

Repro before:
1. Create a file such as `bindings.ts` with:

   const constantValue = "abc"; let mutableValue = "def"; var legacyValue = "ghi";

   const functionExpr = function () { return 1; }; const arrowFn = () => 2; let generatorExpr = function* () { yield 3; };

2. Run the parser / semantic diff flow against that file.
3. Observe that:
   - `constantValue` is reported as `variable` instead of `constant`
   - function-valued declarations are not consistently classified between parser implementations
   - generator-valued declarations are not surfaced as `generator`

Expected behavior after this change:
- `constantValue` => `constant`
- `mutableValue` => `variable`
- `legacyValue` => `variable`
- `functionExpr` => `function`
- `arrowFn` => `function`
- `generatorExpr` => `generator`

Verification:
- added Vitest coverage for JS/TS binding classification
- added `sem-core` unit coverage for the same cases
- ran `cargo test -p sem-core`

### DIFF
--- a/crates/sem-core/src/parser/plugins/code/entity_extractor.rs
+++ b/crates/sem-core/src/parser/plugins/code/entity_extractor.rs
@@ -81,7 +81,7 @@ fn visit_node(
             let entity_type = if node_type == "decorated_definition" {
                 map_decorated_type(node)
             } else {
-                map_node_type(node_type)
+                map_node_type(node, source)
             };
             let content_str = node_text(node, source);
             let content = content_str.to_string();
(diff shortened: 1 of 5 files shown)


Palanikannan1437 (commented):
Thanks so much @MattCozendey for the PR, here's some comments/questions/nitpicks....

rs545837:
Hey @MattCozendey, just checking in. We merged #35 which touches some of the same area (JS/TS entity extraction). Your PR still adds value with the constant/generator classification. Let us know if you're still interested and we can help with the rebase. Thanks\!

PR #30, opened 2026-03-11
Plaintext mode; JSONC support; and mime detection with `file`

### LINKED ISSUE #29: Plaintext based mode
Hi! I really love using this but it would be really nice if there was a plaintext mode that didn't use Unicode blocks. Probably a similar style to git status with ansii color coded output. On small terminal sizes the pretty mode can be a bit too much and feel cluttered.

Image

taciturnaxolotl:
Closes #29 

I also added support for detecting unknown files via the `file` command, so long as it exists. I also noticed in the process of doing that that JSONC breaks the current parsing so I added a fallback on JSON parse that tries JSONC parsing.

image

### DIFF
--- a/src/utils/mime.ts
+++ b/src/utils/mime.ts
@@ -15,16 +15,6 @@ const MIME_TO_PLUGIN: Record<string, string> = {
   // Markdown
   'text/markdown': 'markdown',
   'text/x-markdown': 'markdown',
-  // Code (TypeScript, JavaScript, Python, Go, Rust)
-  'application/typescript': 'code',
-  'text/typescript': 'code',
-  'text/x-typescript': 'code',
-  'application/javascript': 'code',
-  'text/javascript': 'code',
-  'text/x-python': 'code',
-  'text/x-script.python': 'code',
-  'text/x-go': 'code',
-  'text/x-rustsrc': 'code',
 };
 
 let fileCommandAvailable: boolean | undefined;


src/parser/differ.ts
```diff
@@ -31,19 +31,25 @@ export function computeSemanticDiff(
     let beforeEntities: SemanticEntity[] = [];
     let afterEntities: SemanticEntity[] = [];
 
+    const fallback = registry.getPluginById('fallback')!;
```
rs545837: The `!` non-null assertion here will crash if no fallback plugin is registered. Should guard against it:

```ts
const fallback = registry.getPluginById('fallback');
```

Then check `if (plugin !== fallback && fallback)` before calling `fallback.extractEntities`.

src/utils/mime.ts
```diff
@@ -0,0 +1,40 @@
+import { execFileSync } from 'node:child_process';
+
+// Map MIME types returned by `file --mime-type` to parser plugin IDs
+const MIME_TO_PLUGIN: Record<string, string> = {
+  'application/json': 'json',
```
rs545837: This only covers 9 of the 20+ supported languages. Missing at minimum:

```ts
'text/x-c': 'code',
'text/x-c++src': 'code',
'text/x-java-source': 'code',
'text/x-ruby': 'code',
'text/x-php': 'code',
'text/x-csharp': 'code',
'text/x-shellscript': 'code',
```

If MIME detection only works for a few languages, it gives a false sense of coverage.

src/parser/plugins/json/index.ts
```diff
@@ -8,7 +9,9 @@ export class JsonParserPlugin implements SemanticParserPlugin {
   extensions = ['.json'];
 
   extractEntities(content: string, filePath: string): SemanticEntity[] {
```
rs545837: Since you added JSONC parsing, should `.jsonc` be added to the extensions list? Right now `.jsonc` files still fall to the fallback parser.

src/utils/mime.ts
```diff
@@ -0,0 +1,40 @@
...
+  } catch {
+    fileCommandAvailable = false;
+  }
+  return fileCommandAvailable;
+}
+
+export function detectPluginByMime(absoluteFilePath: string): string | null {
+  if (!isFileCommandAvailable()) return null;
+  try {
+    const mime = execFileSync('file', ['--mime-type', '-b', absoluteFilePath], {
```
rs545837: `getPlugin()` is called for every file in a diff. This shells out to `file` synchronously for every unrecognized file that exists on disk. On repos with many extensionless files (Makefile, Dockerfile, LICENSE, etc.) this could add up. Consider caching the result per path, or at least per MIME type lookup.

src/parser/plugins/json/index.ts
```diff
@@ -8,7 +9,9 @@ export class JsonParserPlugin implements SemanticParserPlugin {
...
-    const parsed = JSON.parse(content);
+    const errors: unknown[] = [];
```
rs545837: The check `errors.length > 0 && parsed === undefined` means partially parsed JSONC with errors will silently succeed. Regular `JSON.parse` would have thrown on malformed `.json` files. This changes strictness for normal JSON. Consider keeping strict mode for `.json` and only using lenient JSONC parsing for `.jsonc` files.

src/parser/differ.ts
taciturnaxolotl: done!

rs545837 (commented):
Thanks for the quick fixes! The fallback nullcheck, JSONC strictness, and `.jsonc` extension all look good now.

On the MIME detection question — I'd say keep it, but **only for structured data formats** (JSON, YAML, TOML, CSV, Markdown). Drop the code MIME mappings (`text/x-python`, `text/x-rustsrc`, etc.) because even though the `code` plugin exists, it picks the tree-sitter grammar based on file extension. An extensionless file mapped to `code` via MIME would fail since the plugin doesn't know whether to parse it as Python, Rust, Go, etc.

So the MIME map would just be:

```ts
const MIME_TO_PLUGIN: Record<string, string> = {
  'application/json': 'json',
  'application/x-yaml': 'yaml',
  'text/x-yaml': 'yaml',
  'text/yaml': 'yaml',
  'application/toml': 'toml',
  'text/csv': 'csv',
  'text/markdown': 'markdown',
  'text/x-markdown': 'markdown',
};
```

This covers the realistic use case (extensionless config files) without the false positives on code. The plain formatter looks clean too, nice addition.

taciturnaxolotl:
sounds good! just pushed that

PR #32, opened 2026-03-12
feat: markdown diff formatting

deftdawg:
I needed markdown output to add to PR comments...  hence markdown diff format.

Same idea as https://github.com/Ataraxy-Labs/sem/pull/30

### DIFF
--- a/src/cli/commands/diff.ts
+++ b/src/cli/commands/diff.ts
@@ -9,7 +9,7 @@ import { SemDatabase } from '../../storage/database.js';
 import { formatTerminal } from '../formatters/terminal.js';
 import { formatJson } from '../formatters/json.js';
 import { formatMarkdown } from '../formatters/markdown.js';
-import { createDefaultRegistry } from '../../parser/plugins/index.js';
+  format?: 'terminal' | 'json' | 'markdown' | 'md';
 import { loadConfig, validateChanges, formatValidationResults } from './validate.js';
 
 export interface DiffOptions {


rs545837 (approved):
Nice addition, this is exactly what we needed for PR comment integration. Clean implementation that follows existing patterns well. A few suggestions to tighten it up:

**Rust/TS consistency**
- The Rust formatter distinguishes cosmetic changes (`structural_change == false`) with `~` instead of `Δ`. The TS version doesn't check `structuralChange` at all, so all modified entities get `Δ`. Would be good to match behavior.
- The `"md"` alias only exists on the Rust side (`"markdown" | "md"`). Worth adding in TS too.

**Diff blocks are disconnected from their entity**
Right now diffs get accumulated and dumped after the table. If a file has 5 modified entities, you see a table with 5 rows, then diff blocks below with no way to tell which entity they belong to. Adding the entity name above each diff block would fix this:

```
**`processPayment`**
```diff
- old line
+ new line
```
```

**Help text updated for commands that don't support markdown yet**
`bin/sem.ts` updates the `--format` description for `status`, `log`, `blame`, `history`, `review`, `watch`, `query`, `label`, and `comment` to say "terminal, json, or markdown". But only `diff` actually handles the markdown format. The other commands would silently fall through to terminal output. Either wire markdown into those commands or keep their help text as-is until it's supported.

**Nice to have: line count column**
Since markdown output will likely end up in PR comments, a line count indicator would help readers gauge change size at a glance:

```
| Δ | function | processPayment | +15/-3 lines |
```

Overall this is good to merge with the consistency fixes. Thanks for the contribution!

PR #39, opened 2026-03-31
feature/single-source-of-truth-for-entity-type

MattCozendey:
adds an union for EntityType, better DX. Also fixes 2 existing linting errors


MattCozendey:
Right. Looseness is helpful, but arguably you get more DX security. I personally find it daunting to not be able to tell a type from a function call. Alas, it's your codebase, so feel free to close if you deem it inappropriate. on a secondary note, there's a Rust rewrite taking place, correct? Should I bother patching Typescript?

PR #55, opened 2026-04-06
Feat: add Dart language support 

td2thinh:
## Summary
  - Add tree-sitter-dart https://crates.io/crates/tree-sitter-dart
  - Extract classes, mixins, extensions, enums, typedefs, top-level functions, and class members (methods, getters, setters, constructors, fields, operators)
  - An attempt to handle top level functions grammar of Dart as the AST node for top level functions (including getters and setters) are different from class level functions/methods
 

## Details
- Class members are wrapped in a class_member node (which has no name field), requiring custom name extraction that walks through method_signature/declaration → inner signature → name field
- Top-level functions split function_signature and function_body as bare siblings under source_file, so sibling_function_body() extends the entity content to cover both
- New entity type mappings: mixin_declaration → mixin, extension_declaration → extension, type_alias → type, getter_signature → getter, setter_signature → setter

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,14 +2,53 @@ name: Release
 
 on:
   push:
-    tags:
-      - "v*"
+    branches:
+      - main
+    paths:
+      - "crates/sem-core/Cargo.toml"
 
 permissions:
   contents: write
 
 jobs:
+  check-version:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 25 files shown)


rs545837 (commented):
Nice work on the Dart support! Thorough coverage of constructors, mixins, extensions, fields. A few small suggestions below.

PR #70, opened 2026-04-10
feat: Promote JS/TS const function expressions to functions

MattCozendey:
Classify const declarations with function-like initializers as function entities in the code extractor. Add focused tests covering const promotion and let remaining variable-typed.

This is a rust version of a PR I had opened a few weeks back.


rs545837:
Thanks for opening the PR.

MattCozendey:
You're welcome

MattCozendey:
(I rebased it off latest main)

PR #71, opened 2026-04-10
critical fix(sem-core): detect JS/TS named generator declarations in entity diffs

MattCozendey:
• This fixes a gap in JS/TS entity extraction where named generator declarations were invisible to semantic diff.

Before this change, function* foo() {} and async function* foo() {} were not recognized as entities in JavaScript, TypeScript, or TSX, so additions, deletions, and modifications to those declarations could be missed by sem diff. This PR adds generator_function_declaration to the JS/TS extractor configs, maps it to the existing function entity type, and adds regression tests covering both JS and TS generator declarations.

Tests:

cargo test -p sem-core generator_function_entity_extraction
cargo test -p sem-core nested_functions_typescript


MattCozendey:
Rebased off latest main and resolved conflicts.

PR #72, opened 2026-04-10
feat: add npm wrapper package (requires NPM_TOKEN secret)

MattCozendey:
Creates an npm package wrapper around sem so that users may install into their javascript/typescript/tsx projects and rely on sem without each dev having to individually install. May be useful for ci and/or precommit hooks and/or agent rules where a user may want the agent to be aware of differences more simply

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,6 +110,7 @@ jobs:
         shell: pwsh
         run: |
           Compress-Archive -Path target/${{ matrix.target }}/release/sem.exe -DestinationPath ${{ matrix.name }}.zip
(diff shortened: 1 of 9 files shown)


MattCozendey:
Mostly added this because I want to the dependency on sem be explicit per codebase

MattCozendey:
Thanks for flagging. I checked the full workflow, and release is an existing job id in .github/workflows/release.yml. publish-npm depends on it intentionally because the npm package’s postinstall downloads binaries from the GitHub Release, so the npm publish should wait until the release assets exist. publish-crates only needs create-tag, but npm has that extra release-asset dependency.

PR #74, opened 2026-04-10
Add graph-backed sem grep command and document it across the CLI docs

MattCozendey:
## Summary

This PR adds `sem grep`, a graph-backed semantic search command for finding entities before running impact analysis, narrowing review scope, or doing migrations. It's also supported via the MCP side.

Instead of introducing a separate query language, `sem grep` reuses the existing entity graph and cache, so search results can be filtered using dependency relationships and graph metadata that already exist in `sem`.

## What changed

- added a new `sem grep` command to `sem-cli`
- wired the command into the main CLI surface
- support name matching by default, with optional entity body search via `--content`
- support filters for:
  - `--type`
  - `--path`
  - `--tests`
  - `--depends-on`
  - `--ref-kind`
  - `--min-dependencies`
  - `--min-dependents`
  - `--file-exts`
  - `--no-cache`
- added both terminal output and machine-readable JSON output
- added focused unit tests covering:
  - name vs content matching
  - dependency filtering and `--ref-kind`
  - test-only and graph-threshold filtering
- updated the README and docs site to document `sem grep`, add usage examples, and place it before `sem blame` for consistent command ordering

## Testing

- `cargo test -p sem-cli grep_`

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,30 @@ echo '[{"filePath":"src/main.rs","status":"modified","beforeContent":"...","afte
 sem diff --file-exts .py .rs
 ```
 
+### sem grep
+
+Graph-backed entity search for discovery, migrations, and review narrowing.
+
+```bash
+# Find entities by name
+sem grep auth
+
+# Search inside entity bodies too
+sem grep exec --content
+
+# Filter by entity type and path
+sem grep authenticate --type function --path src/auth
+
+# Find direct callers of an entity
+sem grep --depends-on exec --ref-kind calls
+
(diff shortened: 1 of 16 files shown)


MattCozendey:
that is incorrect. crates/sem-mcp/src/server.rs:24 already has use crate::tools::*;, and crates/sem-mcp/src/tools.rs:48 defines pub struct GrepParams, so GrepParams is in scope through the glob import.

MattCozendey:
Closing as mantainer decided against it on X/Twitter

PR #79, opened 2026-04-12
bugfix/agent-safe-directory-handling

MattCozendey:
## Summary

Fix `sem` repository discovery when running under agent-launched Git environments that provide `safe.directory` through command-line Git config, and apply the same behavior consistently to related git-backed commands.

## Problem

Running `sem diff` from an agent could fail even though the same command worked in a normal terminal. In the agent environment, Git sees `safe.directory` via `GIT_CONFIG_COUNT`, `GIT_CONFIG_KEY_*`, and `GIT_CONFIG_VALUE_*`, but libgit2 owner validation could still reject the repository.

Before this change, `sem diff` also mapped all `GitBridge::open` failures to “Not inside a Git repository,” which masked the real owner-validation error and made the failure look like bad repo discovery rather than a `safe.directory` mismatch.

## Fix

- Preserve non-`NotFound` libgit2 errors in `sem diff`, so owner-validation failures are reported as real git errors instead of “not a repo.”
- Teach `GitBridge::open` to recognize command-line `safe.directory` entries from the `GIT_CONFIG_*` environment variables.
- Retry repository discovery only for libgit2 `ErrorCode::Owner` errors, and only when the nearest git root matches a configured `safe.directory` entry or `*`.
- Guard the temporary libgit2 owner-validation disable behind a process-wide lock and an RAII restore guard, since that libgit2 option is process-global.
- Route CLI `sem blame` and MCP `sem_blame` through `GitBridge` instead of direct `Repository::discover`, so the same safe-directory behavior is centralized and shared.

## Validation

Ran:

- `cargo test -p sem-core`
- `cargo test -p sem-cli`
- `cargo test -p sem-mcp`
- `cargo run -q -p sem-cli -- diff --format plain --file-exts rs`

### DIFF
--- a/crates/sem-cli/src/commands/blame.rs
+++ b/crates/sem-cli/src/commands/blame.rs
@@ -1,7 +1,7 @@
 use std::path::Path;
 
 use colored::Colorize;
-use git2::Repository;
+use sem_core::git::bridge::GitBridge;
 use sem_core::parser::plugins::create_default_registry;
 
 use super::truncate_str;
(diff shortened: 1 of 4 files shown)


MattCozendey:
This is a fair observation, but I don’t think it’s a security bug under our current threat model. GIT_CONFIG_COUNT / GIT_CONFIG_KEY_* / GIT_CONFIG_VALUE_* are Git’s command-line config transport, and this path is intentionally emulating git -c safe.directory=... so agent-launched sem behaves like Git. If an attacker can control the process environment, they can influence command-scoped Git config generally.

MattCozendey:
1 - See message from above(I know you're a bot, just want it on the record).
2 - the drop-order explanation is wrong as written. In /crates/sem-core/src/git/bridge.rs:31, _guard is created before _owner_validation, so Rust drops _owner_validation first and then _guard; that means validation is restored while the mutex is still held, matching the comment.

MattCozendey:
this is blocking my AI agents from fully using sem. Please take a look when you can

rs545837:
Yeah goona review this soon. Thanks for opening the PR.

rs545837:
Thanks for the spotting this interesting scenario of repo ownership, it definitely helps agents haha.

PR #82, opened 2026-04-15
docs: update language count and table to include Scala

sideeffffect:
Documentation-only follow-up to #80.

**README.md**
- Language badge: 23 → 24
- "23 programming languages" → "24 programming languages"
- Add Scala row to the language table (after Kotlin, as both are JVM languages):
  - Extensions: `.scala` `.sc` `.sbt` `.mill` `.kojo`
  - Entities: classes, objects, traits, enums, functions, given instances, extensions, type aliases

**SKILL.md**
- "23 language plugins" → "24 language plugins"


rs545837:
WIll update it once I add the zig support as well. 

Thanks for opening the PR, I will push the change soon.

PR #84, opened 2026-04-15
add zig support

df51d:
Will try to make it better for a while, hence the draft.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1369,6 +1369,7 @@ dependencies = [
  "tree-sitter-swift",
  "tree-sitter-typescript",
  "tree-sitter-xml",
+ "tree-sitter-zig",
  "xxhash-rust",
 ]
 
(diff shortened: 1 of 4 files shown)


df51d:
Ok I am throwing a towel with improving it further. I guess it will improve over time if needed.
EDIT:
There is bug when commenting out/removing the test, it will show every variable inside as deleted. And I am unsure how to model `SuppressedNestedEntity` so it will work for example for following (var in test)
```  
(test_declaration [69, 0] - [79, 1]
    (string [69, 5] - [69, 38]
      (string_content [69, 6] - [69, 37]))
    (block [69, 39] - [79, 1]
      (variable_declaration [70, 4] - ```

rs545837:
Merged with a few fixes:

- **Entity type promotion**: `const Point = struct { ... }` now correctly shows as `struct` instead of `variable`. Same for enums and unions. Zig defines these via const assignment, so added `promote_zig_variable()` to detect the RHS type.
- **Cleaned up config**: Removed `struct_declaration`, `enum_declaration`, etc. from `entity_node_types` since they never match at the top level in Zig (they're expressions nested inside `variable_declaration`).
- **Fixed test**: Indentation, removed debug print, added type assertions.

Thanks for the contribution!

PR #88, opened 2026-04-17
feature/sem-entities-path-based

MattCozendey:
sem entities now works via path, instead of being scoped to single files. this allows for repo-wide entity search across multiple files at once. The old calling shape remains the same, because "sem entities ./foo.ts" is, after all, also a path. By default, "sem entities" list all entities.

If only one file is found, no grouping happens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,12 +169,17 @@ sem log authenticateUser --json
 
 ### sem entities
 
-List all entities in a file with their types and line ranges.
+List all entities under a file or directory path. No path is the same as `.`.
 
 ```bash
+sem entities
+
+sem entities .
+
 sem entities src/auth.ts
 
 # JSON output
+sem entities --json
 sem entities src/auth.ts --json
 ```
 
--- a/crates/sem-cli/src/commands/entities.rs
+++ b/crates/sem-cli/src/commands/entities.rs
@@ -1,76 +1,224 @@
-use std::path::Path;
+use std::collections::BTreeSet;
+use std::path::{Path, PathBuf};
 
 use colored::Colorize;
+use sem_core::model::entity::SemanticEntity;
 use sem_core::parser::plugins::create_default_registry;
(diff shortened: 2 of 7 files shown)


MattCozendey:
these risks are fair @rs545837 do you like the general direction of this PR? Should I fix those?

rs545837:
yeah looks great, and sounds like a good idea to me.

MattCozendey:
cool! I'll update it soon, thanks

MattCozendey:
Ignore these changes (for now) - i was just rebasing off main.

MattCozendey:
Thanks, agreed on the silent partial-results issue. I updated both sem entities and sem_entities so directory traversal/read failures are surfaced instead of being dropped:

CLI no longer uses walker.flatten() in the entities command; walker errors now fail visibly.
CLI directory extraction now routes each file through the same read/parsing path as file mode, so unreadable files error instead of disappearing from the result.
MCP directory mode now propagates directory traversal errors and file read errors instead of ignoring them.

On the file_path -> path point: I’m intentionally treating this as a breaking change for sem_entities. This project is still new enough that I’d rather make the API shape correct than preserve the old parameter name. To avoid the dangerous fallback case, I added deny_unknown_fields to EntitiesParams, so callers still sending file_path now fail parameter deserialization instead of silently defaulting to "." and scanning the repo.

@rs545837

MattCozendey:
@rs545837 please let me know if you want/need me to change anything

rs545837:
just on a side note, I would greatly appreciate if the comments are written in a more humanly way so that it gets easier to understand the intent 😅.

MattCozendey:
sure.

I said that I understand it's a breaking API change but it's worth it because the project is that new. 😊

MattCozendey:
Again, for the record, that is expected. I fully intend on the breaking api change. This bot really should take comments into consideration.

rs545837:
Thanks for the PR! Merged it

Issue #81, opened 2026-04-15
Typescript - duplicated differences shown for modified class with modified methods within


quincycs:
Hi,

I'm seeing the tool output unnecessary differences twice for my code structure.  One diff segment is showing that the class is modified and shows all the lines different b/c the class is modified.  Another diff segment is showing that the method (within the class) is modified and shows all the lines different b/c the method is modified.  This results in the same lines being printed twice ( one for each diff segment ).  I'm reading the code and I think to myself...didn't i just read this?  Oh yes, I did.

CLI command I'm using

`sem diff -v --from <some-hash> --to <another-hash>`

My code structure.

```
export default class DataStack extends cdk.Stack {
  public PrimaryRdsSg: ec2.ISecurityGroup;
  private Vpc: ec2.Vpc;

  constructor(scope: Construct, id: string, props: DataStackProps) {
    super(scope, id, props);
    // some code
  }

  private genPg18Infra(params: {
    vpcDns: r53.PrivateHostedZone;
    dbSecret: secrets.ISecret;
  }) {
         // some method
  }
```

rs545837:
Hey Quincy, thanks for reporting this. You're right, when a method inside a class changes, sem was showing the diff twice, once for the class and once for the method. The class entity's body includes all its methods, so any method change also registers as a class change.                        

Fixed in the latest push. Now when a method is modified, we skip the parent class from the diff output since the method diff is more specific.

quincycs:
Thank you!  Confirmed it's working

rs545837:
Lemme know if you hit any other issues.

PR #94, opened 2026-04-26
Improve parent dedup: handle renamed/added/deleted children + show parent context in output

nminev:
## Summary

This PR builds on the dedup work in #81 with two improvements:

**1. Broader suppression — handles Renamed, Added, and Deleted children (not just Modified)**

The current implementation in `a03f7e4` only suppresses a parent `Modified` when a child is also `Modified`. It misses the case where a child is *renamed* — which is arguably the most common scenario (renaming a method inside a class). In that case the child shows as `Renamed` but the parent still shows as `Modified`, which is still noise.

The new approach strips child content from both sides of the parent using entity line numbers, then compares what remains. If the parent's own declaration/signature is unchanged, the `Modified` is suppressed regardless of what type of change the children had.

```
// Before (only catches Modified+Modified):
if modified_ids.contains(entity.id) && modified_ids.contains(parent_id) {
    suppress parent
}

// After (catches any child change type):
before_own = parent_content with child lines stripped
after_own  = parent_content with child lines stripped
if before_own == after_own { suppress parent }
```

The parent is **not** suppressed if its own declaration changed (e.g. a class gains a base class), even if children also changed.

**2. Parent context in terminal output**

Nested entities now display as `Parent::entity_name` so you can immediately tell which class a changed method belongs to. Works alongside the existing `old -> new` rename display.

```
// Before:
│  ↻ function   get_card_1                [renamed]

// After:
│  ↻ function   CardService::get_card → …  [renamed]
```

Adds `parent_name: Option<String>` to `SemanticChange` (serialised as `parentName` in JSON, skipped when null — fully additive).

## Test plan

- [ ] All existing tests pass (verified: 185 tests, 0 failures)
- [ ] Updated `test_parent_child_dedup_class_method` to use multi-line content with consistent line numbers (the original single-line fixture was incompatible with line-number-based stripping)
- [ ] New tests: child renamed → parent suppressed; child added → parent suppressed; class declaration changed → parent NOT suppressed; `parent_name` populated correctly; all-children-deleted → parent still shown

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rs545837:
Thanks for mentioning that it's generated with claude code, I will review it.

rs545837:
The suppress_redundant_parents (line 148) already checks if ANY child has a change (changed_ids.contains(entity.id.as_str())), regardless of whether it's Modified, Renamed, Added, or Deleted. 

Then it suppresses the parent if the parent is Modified. This already covers all the cases the PR claims to fix.                                                                              
                                                                  
I think the PR was based on an older version of the code (before the fix in ee94c9e).

rs545837:
Closing it for now, reopen if there's a case that I missed.

nminev:
Hey, fair point on the suppression — you're right that `suppress_redundant_parents` already covers renamed/added/deleted children since it's keyed on `changed_ids` regardless of change type. That part overlaps.

Two things though:

**1. The precision gap**

Your approach suppresses the parent whenever any child has a change. But consider the case where a parent's own declaration changes *at the same time* as a child change — e.g. a class gains a base class *and* a method gets renamed in the same commit. Your code would suppress the parent `Modified` entirely, even though the class declaration itself genuinely changed. The line-number-based stripping approach only suppresses when the parent's own content (with child lines removed) is actually the same before and after, so it handles that case correctly.

It's an edge case, but it's real and the current behavior would silently drop a meaningful change.

**2. The `parent_name` context — this was actually the main thing**

This part wasn't addressed at all in the close. Right now the output shows:

```
│  ↻ method     get_card_1              [renamed]
```

With the change it shows:

```
│  ↻ method     CardService::get_card_… [renamed]
```

When you have multiple classes in a file with similarly named methods, the current output gives you no idea which class the change is in without going and checking the file. The `parent_name` field is fully additive (serialized as `parentName` in JSON, `null` for top-level entities) so it doesn't break anything.

Feel free to keep it closed if neither of these is something you want, just wanted to make sure both points were on the table. Happy to split into two separate PRs if that's easier to review.

nminev:
Opened two smaller PRs instead:

- #97 — the parent name display only (2 files, purely additive)
- #98 — the suppression precision fix (2 files, no data model changes)

PR #97, opened 2026-04-28
Show parent name in terminal output (Class::method_name)

nminev:
This is a follow-up to #94 which was closed. Splitting that PR into two smaller ones — this covers the output display change only.

## What

Nested entities now display as `Parent::name` in the terminal formatter:

```
// before
│  ↻ method     get_card_1              [renamed]

// after
│  ↻ method     CardService::get_card…  [renamed]
```

When a file has multiple classes with similarly-named methods, the current output gives you no way to tell which class a change belongs to without going to the file. This adds that context directly to the diff output.

## Changes

- `SemanticChange` gets a new `parent_name: Option<String>` field (serialised as `parentName` in JSON, omitted when null — fully additive, `#[non_exhaustive]` already covers this)
- `make_change()` in `identity.rs` populates it from `entity.parent_id`
- Terminal formatter uses it to prefix the name

Also fixes a latent UTF-8 panic in the truncation path — the old code sliced by byte index which would panic on non-ASCII names. Replaced with `char_indices`.

## Tests

Existing tests all pass. The field is exercised end-to-end through the existing formatter output.

### DIFF
--- a/crates/sem-cli/src/formatters/terminal.rs
+++ b/crates/sem-cli/src/formatters/terminal.rs
@@ -104,12 +104,21 @@ pub fn format_terminal(result: &DiffResult, verbose: bool) -> String {
             };
 
             let type_label = format!("{:<10}", change.entity_type);
-            let name_display = if let Some(ref old_name) = change.old_entity_name {
+            let base_name = if let Some(ref old_name) = change.old_entity_name {
                 format!("{old_name} -> {}", change.entity_name)
             } else {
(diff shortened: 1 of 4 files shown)


rs545837:
Thanks for opening the PRs, sorry got a bit packed with schedule, will review this today.

rs545837:
Thanks for the PR, it really does make sem more easier to view the diffs, and for the UTF-8 fix.

Issue #78, opened 2026-04-12
Discussion about integrating sem into diff-so-fancy


scottchiefbaker:
I like the concept of `sem` so I took a run at seeing what it would take to integrate it into d-s-f. It ended up being pretty simple. Check out the `sem_integration` branch: https://github.com/so-fancy/diff-so-fancy/tree/sem_integration.

<img width="806" height="945" alt="Image" src="https://github.com/user-attachments/assets/6365e90b-e0fb-44c6-9171-dc85cdc3536a" />

Definitely still needs some work, but I was pretty happy with it for 27 minutes of coding.

rs545837:
Tested the branch, some improvements I think. 

- on line 1373, I think it should be ```$? >> 8```
- sem diff $file always diffs against HEAD, so piping git diff HEAD~3..HEAD shows the wrong sem info. Quick fix on your side would be checking for env vars like SEM_FROM/SEM_TO and passing them as --from/--to to sem. On our side we'll add a --patch mode that reads unified diff from stdin, so you can just pipe each file's diff chunk through instead of calling sem separately.
- Spawning sem diff per file is slow on big diffs. Could call it once without a file filter and parse the output.

Excited for the integration, do lemme know any feedback or if you hit any walls.

nminev:
Adjacent to this integration discussion — I've been maintaining a fork at https://github.com/nminev/sem with a few changes that might be relevant to how this renders in d-s-f, since both come up in the diff summary:

1. **Parent suppression for nested entities.** When a method inside a class changes, only the leaf appears with a `parent::child` path (e.g. `User::validate modified`) instead of two separate entries (`User modified` + `validate modified`). For diff summary headers this halves the line count on the common case.

2. **Recursive JSON entity extraction.** Upstream's JSON parser is depth-2; ours recurses fully. So a `package.json` edit surfaces `scripts::build` rather than just `scripts`. This improves usefulness for config-heavy diffs (which I run into a lot).

3. **Full ancestor chain in `parent_name`.** A new field on each change carrying the joined chain (e.g. `jest::config` for an entry at `/jest/config/testTimeout`), so consumers can render full paths without traversing `parent_id` themselves.

[...]

PR #101, opened 2026-04-30
JSON: recursive entity extraction with suppress-parent display

nminev:
## Why

Today the JSON parser is depth-2: it surfaces top-level keys and one layer of children, then treats everything deeper as opaque text. A `package.json` edit to `scripts.build` reports as `scripts modified` with the entire scripts object as the before/after content — no signal about which script actually changed. For agent and CI consumers parsing the JSON output, this is essentially noise; in the diff-so-fancy integration discussed in #78 it's the difference between "a script changed" and "build was changed."

Builds on the `parent_name` field merged in #97 and the precision guard merged in #98. Rebased onto current main after both landed.

## What changes

**Parser**
- Recurse into every nested object; every key at every depth becomes an entity.
- JSON Pointer (`/scripts/build`, `/jest/config/testTimeout`) is the stable identity.
- Arrays stay opaque (elements have no stable identity), but array-typed keys are still entities and rename-detect via structural hash.
- Entity ID format: `{file}::{pointer}` instead of `{file}::{type}::{pointer}`. A key whose value type changes (scalar ↔ object ↔ array) keeps the same ID and matches Phase 1 as Modified instead of Deleted + Added. This is the only breaking output change.

**`parent_name` traversal**
- Builds the full ancestor chain by walking `parent_id` (e.g. `jest::config` for an entity at `/jest/config/testTimeout`), instead of returning only the leaf segment of the immediate parent's ID.
- Skips empty ancestor names — package-lock.json uses `""` as the root-project key, which previously rendered as `packages::::dependencies` (double colon). The full pointer is still recoverable from `entity_id`; only the displayed chain drops the empty segment.

**Suppression extends #98's precision check**
- Adds `\"object\"` to `CONTAINER_TYPES` so a JSON parent isn't reported alongside a child change of its own. With recursion this matters more than before — without it, every leaf change cascades a chain of parent Modifieds.
- The Modified-suppression branch additionally requires `entity_type` to match across before/after, so scalar↔object value-type transitions keep the parent change (the type change itself is meaningful).
- An additional pass drops Moved entries when the entity's `old_parent_id` is in the change set. This catches the case where a parent rename (e.g. `scripts → tasks`) couldn't be matched as Renamed because its children also changed, but the children themselves matched by structural hash and surfaced as Moved. Without this, the rename-failure surfaces both the parent Deleted/Added and the child Moves; with it, only the leaf Moves remain — and consumers can infer the parent rename from the children's `old_parent_id` + new `parent_name`.

**Spec doc**
- `JSON_SEMANTIC_DIFF_SPEC.md` documents the behavior end-to-end: entity extraction rules, change cases, display format, the matching algorithm overview, and known limitations (chiefly the parent-rename + content change case).

## Tests

38 BDD-style tests in `json.rs` covering top-level changes, nested rename/add/delete, deep nesting, type transitions, array opacity, parent-rename fallbacks, document edge cases, and the empty-string-key case from package-lock.json. They use `compute_semantic_diff` (the public pipeline) so they assert observable behavior rather than internal state.

All 190 sem-core tests pass (188 pre-existing + 38 new − 36 reorganized).

## Notes

- The behavioral change to suppress JSON object parents directly contradicts the previous spec language describing double-reporting as "intentional, both useful." For non-JSON code with depth-1 parent-child structures this is unchanged; the new behavior only kicks in for JSON's recursive depth where double-reporting cascades into 3+ entries per logical change. Open to any framing the maintainers prefer here, including splitting suppression off into a follow-up PR if you'd rather merge the parser improvement first.
- Discussed in #78 in the context of the diff-so-fancy integration — the behaviors here are most visible in summary headers showing many small changes.

## Test plan

- [ ] CI green
- [ ] Spot-check with a real package.json / tsconfig.json / pnpm-lock.yaml diff to confirm output looks right
- [ ] Confirm non-JSON entity extraction is untouched (190 sem-core tests pass)


rs545837:
could you rebase the PR to main?

nminev:
Just rebased onto current main — brings in #98's precision guard. Force-pushed `a3b96b3`.

The Modified-suppression path now uses `strip_children_content` from #98; the only addition there is an `entity_type` mismatch check so a JSON value-type transition (scalar ↔ object) keeps the parent change instead of being suppressed. The other two passes (suppress an old parent left behind by a Moved child; drop "stale" Moved entries when the parent renamed) are layered on top of the existing logic.

Re: the inspect bot findings on `extract_object_value` — both are valid (a stray closing `]` could terminate extraction early at depth 0, and the `depth -= 1` on `usize` is a panic risk on malformed input). Happy to push a small fix in this PR or split it out as a follow-up — let me know which you prefer.

rs545837:
Thanks a lot for the thorough work here, the deep entity extraction for JSON is a clear improvement over the depth-2 approach.                                                                                           
                                                      
One concern: we recently converted all AST walkers from recursive to iterative (b9384a0, fixing #103) to avoid stack overflows on deeply nested trees. This PR reintroduces recursion for JSON entity extraction, which could hit the same issue on files like package-lock.json or pnpm-lock.yaml that can nest hundreds of levels deep.

Could you convert the recursive traversal to an iterative worklist to stay consistent with the rest of the codebase?

Also I know that the PR might have some long code, so would appreciate if there are things that can be cut down. 
I am planning to remove some wordy code here and there in sem, which can help everyone maintain it.

nminev:
Pushed `e8592d3` on top — iterative conversion plus the trim pass. Fast-forward, no force-push.

`extract_entries_recursive` is now `extract_entries` with a `Vec<Frame>` worklist (matches the pattern in b9384a0). Each frame carries a `cursor` through its entries; encountering an object child pushes the parent frame back with the cursor advanced and the child frame on top, then `break`s — so the child is popped next, processed fully, and the parent resumes at the next sibling. DFS pre-order is preserved bit-for-bit, so `match_entities` sees the same entity Vec ordering as before.

The two inspect-bot findings on `extract_object_value` are addressed in the same commit: the slice is now returned only on a closing `}` at depth 0 (a stray `]` no longer terminates early), and `depth -= 1` is replaced with `saturating_sub` on both arms. Added `malformed_input_does_not_panic` that runs seven malformed inputs (`{`, `{"a":`, `{"a": {`, `{"a": {] }}`, `{"a": {"b": [}]}`, `{"a": }}}}`, `{"a": {"b": 1}, "c":`) through the public API to lock in the safety property — would catch a regression if `saturating_sub` is removed.

Trim pass: dropped `JSON_SEMANTIC_DIFF_SPEC.md` and consolidated some redundant tests (three pointer-escape tests into one table-driven, two scalar↔array transitions into one, three root-document-edge tests into one). Net is -459 lines across the two files.

On the spec doc — I removed it to cut down the MR but happy to bring it back, either as-is or in a trimmed form, whichever you prefer. Your call.

186 sem-core tests pass.

rs545837:
Thanks a lot for this PR @nminev, appreciate the effort.

PR #102, opened 2026-04-30
skip invalid files on path for entities

Celsiusss:
When you run `sem context` on a directory, the whole command (or mcp command) fails if it contains any binary files. This fixes that issue. It still fails on single files.

Also my rust fmt is going haywire


rs545837:
Thanks for the fix! Binary files crashing the whole command was a feedback I was receiving this week.

PR #110, opened 2026-05-11
Update tree-sitter-dart to 0.2.0

kevmoo:
- Bumps `tree-sitter-dart` dependency from 0.1.0 to 0.2.0 in `Cargo.toml`.
- Updates AST entity extraction in `entity_extractor.rs` to handle the new `method_declaration` node that was introduced in the updated parser.
- Modifies comments in tests to accurately reflect the behavior of `tree-sitter-dart` 0.2.0.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1826,9 +1826,9 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-dart"
-version = "0.1.0"
+version = "0.2.0"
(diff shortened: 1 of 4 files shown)


rs545837:
thanks again for a third PR on dart lang support. Really appreciate it.

kevmoo:
@rs545837 SUPER excited to use your tool more in our ecosystem! 🙏

Issue #105, opened 2026-05-05
Slow performance on Go monorepos: scope-resolve runs sequentially and graph cache never persists


enilsen16:
## Summary

Running `sem context` or `sem impact` on a Go-only repo with ~1,700 files takes **4–7 minutes per call, single-threaded**, and a fresh cache stays at 24 KB regardless of how many runs complete successfully. The same operations on a similarly-sized Elixir codebase return in ~2 seconds at >300% CPU. This makes `sem` unusable for any per-entity sweep workflow on Go monorepos (e.g., piping `sem diff --format json | jq` and looping `sem context --entity-id` over each result).

`sem` version: 0.5 (Homebrew install).

## Reproduction

On any Go-only repo with ~1,500+ `.go` files (no vendor dir):

```
# 1. Get an entity id from sem diff
sem diff --from main --to refs/pr/<N> --format json | jq -r '.changes[] | select(.entityType=="function") | .entityId' | head -1
# (e.g. "internal/foo/bar.go::function::DoThing")

# 2. Run sem context on it
time sem context --entity-id "internal/foo/bar.go::function::DoThing" --budget 4000
# Observed: 4:11 wall, 100% CPU (single-threaded)

# 3. Run sem impact
time sem impact --entity-id "internal/foo/bar.go::function::DoThing"
# Observed: 4:46 wall (warm), 6:07 wall (cold), 100% CPU

# 4. Try the various scoping flags
time sem impact --entity-id <ID> --depth 1
# Observed: 5:44 wall — barely helps

time sem impact --entity-id <ID> --dependents
# Observed: 6:48 wall — actually slower

time sem graph --json > /dev/null
# Observed: 3:58 wall, exit 0
ls -lah .sem/cache.db
# Observed: 24K (mtime updated on rebuild but size stays at the schema-only baseline)
```

For comparison, the same shape of test on an Elixir repo with similar file count returns **2.3 s wall, 318% CPU** — clearly multi-threaded and cached.

## Probable root cause

After reading `crates/sem-core/src/parser/`:

[...]

PR #112, opened 2026-05-15
perf: index entity lookups and parallelize serial passes

### LINKED ISSUE #111: `resolve_with_scopes` Pass 1 still serial on large polyglot repos (follow-up to #105)
## Summary

After the work in #105 / 8b0dc99 (cache persistence fix + Pass 2 parallelization, cli/cli 28 s → 4 s), there is still a substantial serial bottleneck on larger polyglot repos. On a ~10,600-file monorepo (Swift app + TypeScript web + supporting services, 203 k entities, 669 k edges, M4 Max 16-core) cold-cache `sem impact` takes **150–260 s wall** before returning any result. Once `.sem/cache.db` is warm, the same query is **~1 s**.

The remaining cost is concentrated in:

- **Pass 1 of `resolve_with_scopes`** (`scope_resolve.rs:185-213`) — still a plain `for` loop over `parsed_files`, even though Pass 2 was converted to `par_iter` in 8b0dc99. Inside that loop, `scan_return_types` and `scan_init_self_attrs` are called per file.
- **`scan_return_types`** (`scope_resolve.rs:1107-1112`) — does `all_entities.iter().find(|e| …)` per function AST node, scanning the entire 200 k-entry global entity vector.
- **`build_scopes_from_ast`** (`scope_resolve.rs:450-454` and `:476-483`) — same linear scan pattern per class-like node, then iterates `all_entities` again to gather each class's children. Called from the parallel Pass 2 but the per-task work is still O(C_file · E_total).
- **`infer_constructor_param_types`** (`scope_resolve.rs:1638`) — another serial `for (_, _, tree) in parsed_files` AST walk.
- **`build_import_table`** (`graph.rs:1219`) — serial `for file_path in file_paths`, re-`read_to_string`s every file on disk a second time (the per-file content was already in memory from the par-iter Pass 1 of `EntityGraph::build`), and applies multiple `regex::Regex` captures per JS/TS/Rust file.

Total work scales roughly as `O(F · Fns_per_file · E_total)` for the global linear scans, with `E_total` ≈ 2 × 10⁵ in this repo. CPU/wall ratio across the unfiltered run is ~2.3× on a 16-core machine — only ~2 cores busy on average, consistent with the serial sections dominating.

`sem` version under test: **0.5.4** (Homebrew, `/opt/homebrew/bin/sem`).

## Why this surfaces as a hang in agent / MCP / CI contexts

The CLI and the MCP server (`crates/sem-mcp`) both pay this cold cost on first use in a checkout. Default tool timeouts in agent harnesses and MCP clients are typically 15–30 s; they fire well before completion and surface as `exit=142` / "tool timed out". There is also no progress output to stderr, so callers can't distinguish "in progress" from "stuck". The combination produces consistent bug reports of the form "`sem impact` hangs" — when in fact it would complete in a few minutes if given the chance.

Two workflow conditions compound this:

**1. Per-feature git worktrees.** `.sem/cache.db` lives in the working directory the CLI was invoked from (`find_supported_files` in `sem-cli/src/commands/graph.rs:56-85` walks from that root with `hidden(true)`). The cache is therefore never shared between worktrees, even when they share the same `.git/` common dir and the same `HEAD`. Any workflow that creates a fresh worktree per branch/PR/agent-driven-task — a pattern that's increasingly the default for AI assistants and CI executors — pays the full cold cost *once per worktree*, not once per repository.

**2. No sem-local ignore mechanism.** Today's walk only honors `.gitignore` / `git_global` / `.git/info/exclude` (`graph.rs:60-65`); `.semrc` only configures *extension → plugin* mappings (`crates/sem-cli/src/commands/mod.rs:17-22`). Files that are deliberately committed but uninteresting for code-impact analysis — generated GraphQL schemas, large i18n bundles, lint-report JSON, vendored snapshots, fixture trees — therefore inevitably enter `all_entities` and inflate every linear scan above. The only ways to exclude them today are (a) `.gitignore` them (wrong — they're real sources), (b) pass `--file-exts` and lose cross-language edges, or (c) edit upstream.

These two interact: 3 worktrees this week × a repo with ~10 % of its committed mass in "boring parseable" files = paying the entire 200 k-entity cold cost three times.

## Reproduction

All runs from the repo root, `.sem/` removed between cold runs, stdout discarded to `/dev/null`, wall time via `/usr/bin/time -p`. The specific entity name and file path are elided as `<entity>` and `<path/File.swift>` — any small Swift utility function reproduces the same shape.

### Cold-cache full run completes, it is not deadlocked

```sh
rm -rf .sem
sem impact <entity> --file <path/File.swift> --json
# real 150.53s   user 348.68s   sys 5.66s
# exit 0, ~1.8 KB of JSON, 416 MB .sem/cache.db written
```

Immediate re-runs hit the cache:

```sh
sem impact <entity> --file <path/File.swift> --json
# real 1.24s    user 0.72s    sys 0.41s
# real 1.35s    user 0.82s    sys 0.41s
```

Independent invocations under load varied between 150 s and 264 s wall but all completed. There is no deadlock and no infinite loop — the 30 s alarms in earlier internal bug reports simply fired before the build finished.

### Repository scale

```
10 607  tracked files (rg --files)
 4 228  .ts + .tsx
 1 406  .json
 1 148  .swift
   373  .sql
   306  .md
   187  .tf
   184  .yaml
   131  .go
   127  .js
    84  .py
    71  .html
    49  .graphql
    …
203 222 entities    (from `sem graph` plain output, cold-built)
669 247 edges
416 MB  .sem/cache.db after a full cold build
```

### Decomposition by file-extension scope (cold cache, `--no-cache`)

| Extensions included                            | Files included | Real (s) | CPU user (s) |
|------------------------------------------------|---------------:|---------:|-------------:|
| `.swift` only                                  |          1 148 |     1.48 |         4.88 |
| `.ts .tsx`                                     |          4 228 |     8.45 |        17.06 |
| `.json` only                                   |          1 406 |     1.07 |         1.17 |
| `.swift .ts .tsx`                              |          5 376 |     9.29 |        26.58 |
| `.swift .ts .tsx .json`                        |          6 782 |    10.74 |        23.78 |
| `.swift .ts .tsx .md`                          |          5 682 |    13.92 |        33.98 |
| `.swift .ts .tsx .py` (+ 84 files)             |          5 460 |    12.70 |        32.12 |
| `.swift .ts .tsx .go` (+ 131 files)            |          5 507 |    43.60 |        91.16 |
| `.swift .ts .tsx .js .py .go .rs`              |          5 718 |    71.17 |       148.03 |
| all code, excluding `.json`                    |          ~9 k  |    82.33 |       236.82 |
| all (everything `ignore` walks + parses)       |         10 607 |   143.83 |       351.03 |

The most informative row is **`.go`**: adding only 131 Go files to a swift+ts+tsx baseline makes cold build **4.7× slower** (9.3 s → 43.6 s). The 131 added files alone parse in well under a second; the extra time is global cross-language work, not per-file parsing. This is the smoking gun for cost scaling with `E_total` rather than `F`.

## Root cause (with source citations)

`sem-cli` `impact` → `super::graph::get_or_build_graph` → on cold cache → `EntityGraph::build` in `crates/sem-core/src/parser/graph.rs:99`.

1. **Pass 1 — parse all files in parallel** (`par_iter`, `graph.rs:106-115`). Fast, not the problem.
2. **Build the global entity / symbol tables** (`graph.rs:127-197`). O(E), cheap.
3. **`build_import_table`** (`graph.rs:201`, defined at `graph.rs:1219`). Serial `for file_path in file_paths` loop that re-`read_to_string`s every file in the repo (`graph.rs:1228`) and applies multiple regex captures per JS/TS/Rust file. Single-threaded I/O + regex work over 10 k files.
4. **`scope_resolve::resolve_with_scopes`** (`graph.rs:211`, defined at `scope_resolve.rs:77`). This is the dominant cost. It contains:

    - **Pass 1 — serial** loop over every parsed file (`scope_resolve.rs:185-213`):

      ```rust
      for (file_path, content, tree) in parsed_files {
          …
          scan_return_types( … all_entities, …);
          scan_init_self_attrs( … all_entities, …);
      }
      ```

      `parsed_files` includes every file whose extension has a `scope_resolve` config — Swift, TS/TSX/JS, Python, Go, Rust, Java, Ruby, C#, PHP, C++, Kotlin, Dart, Scala, Zig, Bash. Nothing in this loop is parallel — 8b0dc99 parallelized Pass 2 only.

    - Inside `scan_return_types` (`scope_resolve.rs:1107-1112`):

      ```rust
      let func_entity = all_entities.iter().find(|e| {
          e.file_path == file_path && e.name == func_name && {
              let line = node.start_position().row + 1;
              e.start_line <= line && line <= e.end_line
          }
      });
      ```

      A linear scan of every entity in the repo, per function AST node, per file. With ~200 k entities and tens of thousands of function nodes this is the leading constant.

    - Inside `build_scopes_from_ast` (`scope_resolve.rs:450-454`):

      ```rust
      let class_entity = all_entities.iter().find(|e| {
          e.file_path == file_path && e.name == class_name
              && matches!(e.entity_type.as_str(),
                          "class" | "struct" | "interface")
      });
      ```

      And again at `scope_resolve.rs:476-483`:

      ```rust
      for entity in all_entities {
          if entity.parent_id.as_ref() == Some(&ce.id) { … }
      }
      ```

      A linear scan per class-like node, then a *second* linear scan of `all_entities` for every class found to gather its children. `build_scopes_from_ast` runs inside the parallel Pass 2 (`scope_resolve.rs:229-364`) but each task is still O(C_file · E_total), so wall-clock benefit from Rayon plateaus once `E_total` grows.

    - **`infer_constructor_param_types`** (`scope_resolve.rs:218`, defined at `scope_resolve.rs:1638`). Another serial `for (_, _, tree) in parsed_files` walk over every AST.

5. **Pass 2 of `EntityGraph::build`** (`graph.rs:220-337`). `par_iter`, fast — dot-chain + bag-of-words resolution per entity using the precomputed tables. Once we are here, the heavy lifting is done.

In short: the cold build's dominant term is `O( F × Fns_per_file × E_total )`, with `E_total` ≈ 2 × 10⁵ on this repo. The Go-only row in the decomposition table above isolates this scaling cleanly.

## Related

- #105 — earlier serial-resolve report on Go monorepos. 8b0dc99 fixed cache persistence and parallelized Pass 2 of `resolve_with_scopes`. This issue is about the residual serial work in Pass 1 / `infer_constructor_param_types` / `build_import_table`, plus the linear `all_entities` scans inside, which become dominant once Pass 2 is parallel and `E_total` grows past ~10⁵.
- #104 — `scan_return_types` stack overflow on deeply nested code, fixed by converting AST walks to iterative. That change is orthogonal to the issue here — the linear `all_entities.iter().find` at `scope_resolve.rs:1107` lives at the leaf of that now-iterative walk and is unaffected by it.

rs545837:
## Summary

Addresses #111 — `resolve_with_scopes` Pass 1 still serial on large polyglot repos.

On a 10,600-file polyglot monorepo (203k entities, 669k edges), cold-cache `sem impact` takes 150–260s. The bottleneck is O(E_total) linear scans in `scan_return_types`, `build_scopes_from_ast`, and `build_import_table`, plus serial loops that don't use available cores.

**Changes:**

- **HashMap indexes** replace linear scans:
  - `entities_by_file` groups entities by file_path — O(1) instead of O(E) per lookup in `scan_return_types` and class entity find
  - `children_by_parent` groups entities by parent_id — O(1) instead of O(E) per class in `build_scopes_from_ast`

- **Parallelize Pass 1** (`scan_return_types` + `scan_init_self_attrs`) with `par_iter` and merge

- **Parallelize `infer_constructor_param_types`** inner loop with `par_iter`

- **Parallelize `build_import_table`** and reuse in-memory file content instead of re-reading every file from disk

**Expected impact:** The HashMap indexing alone should reduce dominant cost from O(F × Fns × E_total) to O(F × Fns). Combined with parallelization, estimated 10-15x speedup on cold builds for repos at this scale.

## Test plan

- [x] All 208 existing tests pass (`cargo test -p sem-core`)
- [ ] Benchmark on reporter's repo or equivalent 10k+ file polyglot repo
- [ ] Verify cache.db output is byte-identical before/after

Closes #111

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,24 @@ node_modules/
 .npm-cache/
 vendor/
 
+# Scratch / demo files
+*.cast
+*.gif
+*.mp4
+*.png
+*.svg
+attention_*.html
+attention_*.py
+feel.html
+sem_log_demo.html
+gen_*.py
+bench/
+bench_*.py
+bench_*.sh
+demo.*
+docs/*.html
+hero.svg
+
 # Nixos artifacts
 .envrc
 .direnv/
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -84,6 +84,7 @@ impl DiskCache {
             }
         }
 
+        // Insert entities with prepared statement (already in a transaction, so fast)
         {
             let mut stmt = tx.prepare(
                 "INSERT OR REPLACE INTO entities (id, name, entity_type, file_path, start_line, end_line, content, content_hash, structural_hash, parent_id, metadata_json) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11)",
@@ -109,6 +110,7 @@ impl DiskCache {
             }
         }
 
+        // Insert edges with prepared statement
         {
             let mut stmt = tx.prepare(
                 "INSERT INTO edges (from_entity, to_entity, ref_type) VALUES (?1, ?2, ?3)",
@@ -140,21 +142,30 @@ impl DiskCache {
             return None;
         }
 
+        // Load all cached mtimes in one query and validate against disk
         let mut stmt = self
             .conn
-            .prepare("SELECT mtime_secs, mtime_nanos FROM files WHERE path = ?1")
+            .prepare("SELECT path, mtime_secs, mtime_nanos FROM files")
             .ok()?;
+        let cached_mtimes: HashMap<String, (i64, i64)> = stmt
+            .query_map([], |row| {
+                Ok((
+                    row.get::<_, String>(0)?,
+                    (row.get::<_, i64>(1)?, row.get::<_, i64>(2)?),
+                ))
+            })
+            .ok()?
+            .filter_map(|r| r.ok())
+            .collect();
+
         for file in files {
+            let (secs, nanos) = cached_mtimes.get(file.as_str()).copied()?;
             let full = root.join(file);
             let meta = std::fs::metadata(&full).ok()?;
             let mtime = meta.modified().ok()?;
             let dur = mtime
                 .duration_since(std::time::UNIX_EPOCH)
                 .unwrap_or_default();
-
-            let (secs, nanos): (i64, i64) = stmt
-                .query_row(params![file], |row| Ok((row.get(0)?, row.get(1)?)))
-                .ok()?;
             if secs != dur.as_secs() as i64 || nanos != dur.subsec_nanos() as i64 {
                 return None;
             }
--- a/crates/sem-core/src/parser/graph.rs
+++ b/crates/sem-core/src/parser/graph.rs
@@ -198,7 +198,7 @@ impl EntityGraph {
 
         // Build import table: (file_path, imported_name) → target entity ID
         // e.g. ("io_handler.py", "validate") → "core.py::function::validate"
-        let import_table = build_import_table(root, file_paths, &symbol_table, &entity_map);
+        let import_table = build_import_table(root, file_paths, &symbol_table, &entity_map, Some(&parsed_files));
 
         // Run scope-aware resolver for supported languages (reuse pre-parsed trees)
         let has_scope_lang = file_paths.iter().any(|f| {
@@ -575,7 +575,7 @@ impl EntityGraph {
         }
 
         // Build import table from ALL files (imports may reference stale entities)
-        let import_table = build_import_table(root, all_file_paths, &symbol_table, &entity_map);
+        let import_table = build_import_table(root, all_file_paths, &symbol_table, &entity_map, Some(&parsed_files));
 
         // Run scope-aware resolver only on files that need resolution
         let resolve_file_paths: Vec<String> = all_file_paths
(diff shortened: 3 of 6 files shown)


Iron-Ham:
@rs545837 if you'd like me to benchmark a set of changes on my local repo, let me know and I'm happy to do it. Just let me know when the branch is ready for me to pull, build and bench!

rs545837:
Yeah it would be great, if you can benchmark on your repo as well parallely, I am testing on grafana currently.

Iron-Ham:
Awesome, will run locally and get back to you.

Iron-Ham:
Benchmarked this PR against the same ~10,600-file polyglot monorepo from #111 (203,140 entities, 669,108 edges, M4 Max 16-core, Darwin 25.4.0). Same-session A/B between Homebrew `sem 0.5.4` and a `cargo build --release` of `0f23f09`, run back-to-back against the same Git HEAD with `.sem/` removed before every cold trial. Two trials per workload; `sem graph --no-cache` was used everywhere to isolate the algorithmic cost from the SQLite write/read path.

I had claude format the data, as follows:

_____

### Cold-build wall time

| Workload                              | HB 0.5.4 trial 1 | trial 2 | PR #112 trial 1 | trial 2 | Mean speedup |
|---------------------------------------|-----------------:|--------:|----------------:|--------:|-------------:|
| `--file-exts .swift .ts .tsx`         |          20.04 s | 22.17 s |         16.88 s | 17.15 s |        1.24× |
| `--file-exts .swift .ts .tsx .go`     |         110.07 s | 73.12 s |         42.39 s | 39.96 s |        2.23× |
| Full unfiltered                       |         245.18 s |264.01 s |        116.32 s |135.83 s |        2.18× |
| Cold `sem impact` (full + cache write) |         261.42 s |       — |        100.54 s |       — |       2.60× |

The "+Go cliff" — adding 131 Go files to a swift+ts+tsx baseline — shrinks from ~4.3× to ~2.4×, which is exactly the area the indexes and Pass-1 parallelization target. Entity and edge counts are identical across binaries on every workload (203,140 / 669,108 unfiltered; 70,510 / 143,324 with the Go set), so no observable correctness change.

### CPU/wall ratio (parallelism utilization, 16 cores available)

| Workload              | HB ratio | PR ratio |
|-----------------------|---------:|---------:|
| swift+ts+tsx          |     1.85 |     2.03 |
| swift+ts+tsx+go       |     1.30 |     2.76 |
| Full unfiltered       |     1.70 |     3.55 |

Average busy cores roughly doubles on the workloads big enough for the parallelization to dominate. That's the Pass-1 / `infer_constructor_param_types` / `build_import_table` par-iter conversions clearly engaged. We're still well under the 16-core ceiling though — at 3.6 avg busy cores, ~78 % of CPUs are idle on average during a cold full build. There's room left, presumably outside the four hotspots this PR addresses (most likely the SQLite cache write and any residual O(E_total) work that escaped the index conversion).

### Variance is meaningfully better

| Workload          | HB range | PR range |
|-------------------|---------:|---------:|
| swift+ts+tsx+go   | 73 – 110 s (50 % spread) | 40 – 42 s (5 % spread) |
| Full unfiltered   | 245 – 264 s (8 % spread) | 100 – 136 s (32 % spread) |

On the Go-cliff workload the PR is dramatically more predictable. Full unfiltered is noisier on PR than HB, but both are bounded — and the worst-case PR time (136 s) is still well below the best-case HB time (245 s), so the ordering is stable across all trials.

### Warm-cache behaviour

Warm `sem impact` against a populated `.sem/cache.db`:

| Cache built by | Reader | Wall (best of 2) |
|----------------|--------|-----------------:|
| HB             | HB     | 1.74 s |
| HB             | PR     | 1.84 s |
| PR             | PR     | 1.86 s |
| PR             | HB     | 1.21 s |

Cache file is 416 MB on both builds; cross-version reads work in both directions. Warm-cache cost is unchanged within noise.

### Note on the PR description's "10–15× speedup" estimate

At this repo scale I measured ~2.2× on full unfiltered cold build and ~2.2× on the +Go cliff — strong but not 10×. Plausibly the 10–15× was the theoretical maximum assuming the four hotspots are ~all the cost. In practice roughly half the cold-build wall time survives the changes.

### Things I did *not* verify

- I did not byte-compare the SQLite cache blob — the logical graph (entity ids, edge tuples) is identical, but storage-level differences (row order, page layout) could exist. 
- Only one repo / one machine. The PR description's larger speedup claim might hold on a repo with a *different* shape (more classes, fewer tree-sitter-heavy files, etc.). Would be good to also see numbers from cli/cli (which #105 used as the canonical Go-heavy benchmark) before and after this PR.

Iron-Ham:
@rs545837 you got it

rs545837:
@Iron-Ham Thanks for the thorough benchmark, incredibly helpful. The "+Go cliff" you identified (131 Go files adding 34s to a 9s baseline) was the smoking gun.

**Root cause found and fixed in `2e34faa`:** The scope resolver's `register_go_package_imports` was iterating the entire symbol table (203K entries) for every Go import in every file. That's `131 files × ~15 imports × 203K = ~400M iterations`, exactly explaining why 131 Go files cost 34s.

The fix delegates Go import resolution to a pre-built package index (O(1) lookup per import) and removes the redundant Go import expansion from `build_import_table` entirely.

On our benchmarks (different machine, Apple Silicon):
| Repo | Before PR | After all 4 commits | Speedup |
|------|-----------|---------------------|---------|
| cli/cli (856 Go) | 64.6s | **1.2s** | 54× |
| terraform (1928 Go) | 661s | **8.2s** | 81× |
| grafana (5955 Go / 21K total) | ... | **99s** | . |

The "+Go cliff" should be essentially eliminated now. Would you be able to re-benchmark with the latest HEAD of this branch? Particularly interested in whether the `.swift .ts .tsx .go` workload now lands close to the `.swift .ts .tsx` baseline (~17s).

Iron-Ham:
Claude summary follows 🤖 
But performance benchmark looks great 

________

Re-benchmarked against `2e34faa` on the same ~10,600-file polyglot monorepo from #111. Short answer to your question: **yes, the `+Go` cliff is gone — adding 131 Go files to the `swift+ts+tsx` baseline now costs ~2%, not ~5×.**

Same-session A/B against HB 0.5.4 and a `cargo build --release` of `2e34faa`. Two trials per workload, `sem graph --no-cache` to isolate the algorithmic cost from the SQLite paths. Machine state had changed since the previous benchmark (HB itself is also faster today than yesterday), which is why the absolute numbers below differ from my earlier comment — apples-to-apples is the in-session A/B.

### The `+Go` cliff (your specific question)

| Workload                          | HB 0.5.4 (mean of 2) | PR `2e34faa` (mean of 2) | Speedup |
|-----------------------------------|---------------------:|-------------------------:|--------:|
| `.swift .ts .tsx`                 |               6.46 s |                   4.16 s |   1.55× |
| `.swift .ts .tsx .go`             |              32.46 s |                   4.26 s |   7.62× |
| **Cliff (`+Go` / base)**          |           **5.02×** |                **1.02×** |         |

The `+Go` workload (4.26 s) now lands within ~2% of the `swift+ts+tsx` baseline (4.16 s). The `register_go_package_imports` blowup is fully neutralized for this repo. CPU usage is also way down on the `+Go` workload — HB used 68 s of user CPU, the patched binary used 17 s, so this is real work being eliminated, not just better parallelism.

### Full unfiltered cold build (the headline)

| Trial | HB 0.5.4 wall | HB CPU (user) | PR `2e34faa` wall | PR CPU (user) |
|-------|--------------:|--------------:|------------------:|--------------:|
| 1     |      103.61 s |      259.16 s |           17.76 s |       94.14 s |
| 2     |      110.49 s |      269.98 s |           18.14 s |       97.74 s |
| 3 (PR only) |         — |             — |           18.32 s |       98.00 s |
| **mean** | **107.05 s** | **264.57 s** |       **18.07 s** |   **96.63 s** |

**5.92× wall-time speedup, with CPU work cut from 265 s to 97 s — a 2.7× reduction in total work done, not just better parallelism.** End-to-end `sem impact` (including the SQLite cache write) clocked at **20.09 s** cold on this repo, down from the 245–264 s I was seeing yesterday. That brings cold-cache MCP/agent integrations comfortably under any reasonable per-call timeout.

### Variance is now negligible

Three consecutive PR trials of the full unfiltered workload: **17.76 s, 18.14 s, 18.32 s** — a 3% spread. The wild HB variance I saw earlier (73→110 s on `+Go`, 245→264 s on full unfiltered) is gone. Predictability is its own win for tool integrations.

### Correctness signals

Entity counts identical across binaries on every workload tested. Edge counts:

| Workload                  | HB              | PR `2e34faa`   | Δ          |
|---------------------------|----------------:|---------------:|-----------:|
| `.swift .ts .tsx`         | 140,465         | 140,465        | 0          |
| `.swift .ts .tsx .go`     | 143,324         | 142,739        | -585 (-0.4%) |
| Full unfiltered           | 669,108         | 668,500        | -608 (-0.09%) |

The 0.4 % edge drop on the `+Go` workload is **deterministic** (identical across all PR trials) and lives entirely in the Go import resolution path that `2e34faa` rewrote. Based on the commit message — "the regex-based import table was expanding ALL entities from each imported Go package into per-file import entries" — the old code was producing some over-eager edges from package-level expansion that the new AST-level resolver does not. So this reads as a correctness improvement, not a regression, but worth a quick spot-check by someone who knows the Go semantics better than I do (e.g., pick a Go import in a small file and confirm the new edge set matches what the language would actually resolve to).

Warm-cache reads are unchanged (PR warm: 1.02 s / 1.01 s; HB warm against PR cache: 0.69 s — cross-version compatible), cache file size identical at 416 MB.

### Three-way summary

| Metric                          | HB 0.5.4 | PR @ `0f23f09` (yesterday) | PR @ `2e34faa` (today) |
|---------------------------------|---------:|---------------------------:|------------------------:|
| Full unfiltered cold (mean)     | 107.05 s |                   117.56 s |                **18.07 s** |
| `+Go` cliff                     |   ~5×    |                      ~2.4× |                  **~1.0×** |
| CPU work, full unfiltered (mean)| 264.57 s |                   417.16 s |                **96.63 s** |
| Trial-to-trial variance (full)  |    8%    |                        32% |                       **3%** |

The first batch of commits got CPU/wall ratio up (better parallelism) but didn't reduce the actual O(n²) work in the Go import path; `2e34faa` is what made the math right. Net result is a **roughly 6× wall-time speedup** for this repo class — much closer to your description's original 10–15× estimate, and the CPU drop says there's more headroom still by going wider.

rs545837:
I hope things are smoother now @Iron-Ham?

Iron-Ham:
Much better, thank you.

PR #215, opened 2026-05-26
Suppress C and C++ block-local variables

### LINKED ISSUE #211: c cpp local vars extracted
## Summary

C/C++ block-local variables (declared inside function bodies) are extracted as top-level `variable` entities with the enclosing function as `parent_id`. JS/TS already suppresses these. C/C++ does not.

```c
int helper(void) { return 1; }
int main(void) {
  int local = helper();   // <- emitted as an entity
  return local;
}
```

`sem entities` returns `local` as a `variable` entity. Function bodies become noisy in the entity listing; diff and graph inherit the noise.

## Severity

P3 — quality-of-output gap. C/C++ users get a flood of irrelevant entities for every local. The classification is technically truthful (it IS a declaration), but block-locals aren't usually what people mean by "entity" in the sem sense.

## Environment

## Reproduction

```bash
tmp=$(mktemp -d) && cd "$tmp"
git init -q

cat > main.c <<'EOF'
int helper(void) { return 1; }
int main(void) {
  int local = helper();
  return local;
}
EOF

sem entities main.c --json
```

## Observed

```json
[
  {"name":"helper","type":"function", ...},
  {"name":"main",  "type":"function", ...},
  {"name":"local", "type":"variable","parent_id":"main.c::function::main", ...}
]
```

## Expected

```json
[
  {"name":"helper","type":"function", ...},
  {"name":"main",  "type":"function", ...}
]
```

Block-locals should be suppressed (same policy as JS/TS).

## Code references

C/C++ configs in [crates/sem-core/src/parser/plugins/code/languages.rs](https://github.com/Ataraxy-Labs/sem/blob/2447cf74f7ac35466ac613b5a7401b1bf87431de/crates/sem-core/src/parser/plugins/code/languages.rs) include `declaration` as a top-level entity kind and recurse into `compound_statement`. JS/TS has explicit local-variable suppression; C/C++ does not.

## Suggested fix

When extracting a `declaration` node, check whether its nearest enclosing entity is a `function` (or method/lambda). If so, skip emission unless the user has explicitly opted in via a future `--include-locals` flag.

## Independently confirmed

Independently surfaced by a parallel audit (their confidence note: medium — may be intended for global declarations).

## Related Recent Fixes

Adjacent to [#152](https://github.com/Ataraxy-Labs/sem/issues/152), fixed by [#164](https://github.com/Ataraxy-Labs/sem/pull/164), which addressed C/C++ declaration classification. This draft is intentionally lower confidence: it asks whether block-local C/C++ declarations should be suppressed the way JS/TS locals are.

Iron-Ham:
## Summary
- suppress C/C++ `declaration` entities when traversing function-like bodies
- preserve top-level globals and function prototypes
- add C and C++ regression coverage for block-local variables

Closes #211

## Test plan
- `cargo test -p sem-core`
- `cargo test -p sem-cli`
- `cargo build -p sem-cli`
- `/Users/zer0/Developer/oss/sem-wt-a149abe4/crates/target/debug/sem entities main.c --json` in `/Users/zer0/Developer/sem-issue-211-validation-a149abe4`
- `/Users/zer0/Developer/oss/sem-wt-a149abe4/crates/target/debug/sem entities main.cpp --json` in `/Users/zer0/Developer/sem-issue-211-validation-a149abe4`

### DIFF
--- a/crates/sem-core/src/parser/plugins/code/languages.rs
+++ b/crates/sem-core/src/parser/plugins/code/languages.rs
@@ -377,6 +377,30 @@ const JS_TS_SCOPE_BOUNDARIES: &[&str] = &[
     "generator_function",
 ];
 
+/// Inside C function bodies, suppress `declaration` nodes so that block-local
+/// variables are not extracted as nested entities. Inner type declarations are
+/// still reached by traversal after the wrapper is skipped.
+const C_SUPPRESSED_NESTED: &[SuppressedNestedEntity] = &[SuppressedNestedEntity {
+    parent_entity_node_type: "function_definition",
+    child_entity_node_type: "declaration",
+}];
+
+/// Inside C++ function-like bodies, suppress `declaration` nodes so that
+/// block-local variables are not extracted as nested entities. Inner type
+/// declarations are still reached by traversal after the wrapper is skipped.
+const CPP_SUPPRESSED_NESTED: &[SuppressedNestedEntity] = &[
+    SuppressedNestedEntity {
+        parent_entity_node_type: "function_definition",
+        child_entity_node_type: "declaration",
+    },
+    SuppressedNestedEntity {
+        parent_entity_node_type: "lambda_expression",
+        child_entity_node_type: "declaration",
+    },
+];
+
+const CPP_SCOPE_BOUNDARIES: &[&str] = &["lambda_expression"];
+
 #[cfg(feature = "lang-typescript")]
 static TYPESCRIPT_CONFIG: LanguageConfig = LanguageConfig {
     id: "typescript",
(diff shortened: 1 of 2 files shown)


crates/sem-core/src/parser/plugins/code/languages.rs
Iron-Ham, line 1: depending on philosophical evolution of sem, this file should perhaps be a composed file of language specific rulesets or maybe one that supports user extension – but i'll leave that to the maintainers to determine.

Iron-Ham:
Addressed inspect's C/C++ config finding by splitting C and C++ suppression/scope-boundary constants: C no longer references `lambda_expression`, while C++ keeps lambda handling.

The later inspect note about a truncated `assert!(` appears to be a false positive. The PR diff and remote head file both contain the completed assertion, and the updated commit passes local `cargo test -p sem-core`; GitHub `semver` and `build-windows` also pass.

PR #226, opened 2026-05-26
Keep MCP server alive after parse errors

### LINKED ISSUE #168: mcp server terminates on malformed json
## Summary

The `sem mcp` server terminates with exit code 0 when it receives a single line of malformed JSON over stdin, instead of replying with a JSON-RPC `-32700` (Parse error) and continuing to serve. This means any flaky or buggy MCP client can kill the session with one bad frame — line buffering hiccup, stray newline, partial write — with no recovery path.

## Severity

P1 — MCP availability / protocol correctness. A correctly-behaving JSON-RPC server should not exit on a single malformed message. The current behavior means subsequent writes from the client raise `BrokenPipeError` and the agent loses access to sem for the rest of the session.

## Environment

- Reproducible cross-platform.

## Reproduction

```bash
printf '%s\n%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"t","version":"1"}}}' \
  'garbage' \
  | sem mcp
```

## Observed

```
ERROR rmcp::transport::async_rw: Error reading from stream: serde error expected value at line 1 column 1
```

Process exits with status `0`. No JSON-RPC error response is emitted. Subsequent stdin writes from the client raise `BrokenPipe`.

## Expected

Per JSON-RPC 2.0 §5.1, a parse failure should yield:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}}
```

…and the server should continue reading the next frame. Exit code on a parse error should be non-zero only if the server is shutting down for other reasons.

## Code references

The error originates from `rmcp::transport::async_rw` per the stderr trace. The server wiring is in [crates/sem-mcp/src/server.rs](https://github.com/Ataraxy-Labs/sem/blob/2447cf74f7ac35466ac613b5a7401b1bf87431de/crates/sem-mcp/src/server.rs); the transport / framing layer is in the `rmcp` dependency. The fix is likely in how sem-mcp wraps the transport — either by handling per-line parse errors as recoverable, or by configuring the transport not to terminate on them.

## Workaround

Clients can avoid triggering this by validating every outgoing frame is well-formed JSON ending in `\n`. There is no server-side workaround today.

Iron-Ham:
## Summary

- Replace the default rmcp stdio tuple transport with a sem-owned resilient transport.
- Emit JSON-RPC `-32700` parse errors with `id: null` for malformed stdin frames and continue reading.
- Preserve compatibility handling for noisy non-MCP notifications and serialize protocol-error writes with normal responses.

Closes #168

## Test plan

- `cargo test -p sem-mcp`
- `cargo test --workspace`
- `cargo clippy -p sem-mcp --no-deps -- -A clippy::new_without_default -A clippy::collapsible_if -D warnings`
- Manual validation in `~/Developer/sem-mcp-issue-168-dummy`: initialize frame, malformed `garbage` frame, then ping frame returns initialize response, JSON-RPC parse error, and ping response.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1445,6 +1445,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
+ "tracing",
  "tracing-subscriber",
 ]
 
--- a/crates/sem-mcp/Cargo.toml
+++ b/crates/sem-mcp/Cargo.toml
@@ -25,6 +25,7 @@ serde_json = "1"
 schemars = "1"
 ignore = "0.4"
 rusqlite = { version = "0.32", features = ["bundled"] }
+tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 openssl = { version = "0.10", features = ["vendored"], optional = true }
 
--- a/crates/sem-mcp/src/lib.rs
+++ b/crates/sem-mcp/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod cache;
 pub mod server;
 pub mod tools;
+mod transport;
 
 use rmcp::ServiceExt;
 
@@ -18,7 +19,8 @@ pub fn run() -> Result<(), Box<dyn std::error::Error>> {
             .init();
 
         let server = server::SemServer::new();
-        let transport = (tokio::io::stdin(), tokio::io::stdout());
+        let transport =
+            transport::ResilientStdioTransport::new(tokio::io::stdin(), tokio::io::stdout());
         let service = server.serve(transport).await?;
         service.waiting().await?;
         Ok(())
--- a/crates/sem-mcp/src/transport.rs
+++ b/crates/sem-mcp/src/transport.rs
@@ -1,0 +1,339 @@
+use std::{future::Future, sync::Arc};
+
+use rmcp::model::{ClientJsonRpcMessage, ServerJsonRpcMessage};
+use rmcp::transport::Transport;
+use rmcp::RoleServer;
+use serde::Serialize;
+use tokio::io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader};
+use tokio::sync::Mutex;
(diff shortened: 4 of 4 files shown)


Iron-Ham:
Addressed the actionable inspect feedback from this review:

- Preserved all JSON numeric request IDs in invalid-request error responses, including large unsigned numbers and fractional JSON numbers.
- Added tests for string IDs, large numeric IDs, fractional numeric IDs, and invalid ID types.

Notes on the other findings:

- The `receive()` concurrency findings are false positives for the `rmcp::transport::Transport` contract: `receive()` takes `&mut self`, so the transport cannot be read concurrently by multiple tasks through the trait.
- The repeated-invalid-frame loop is intentional for a stdio server: as long as the client continues sending recoverable frames, the server replies to invalid frames and continues serving instead of terminating.
- `null`, boolean, array, and object IDs are intentionally mapped to `null` for invalid-request responses because JSON-RPC request IDs are string, number, or null.

Iron-Ham:
Follow-up for the second inspect review:

- Updated `close()` to take the writer and call `shutdown()` outside the mutex guard.
- Tightened notification compatibility filtering so only id-less notifications can be ignored; requests using a `notifications/*` method are no longer discarded.
- Added tests for close behavior and notification filtering.

The remaining write-lock findings are expected: the mutex intentionally serializes stdout writes so JSON-RPC frames cannot interleave, and writes after `close()` correctly fail as `Transport is closed`.

Iron-Ham:
Final follow-up on the latest inspect pass:

- Tightened `close()` ordering so the writer remains under the mutex until `shutdown()` completes, then it is removed from the transport.

The remaining write-failure behavior is intentional: if stdout cannot accept the JSON-RPC error response, the stdio transport is broken and `receive()` returning `None` lets `rmcp` terminate the service instead of continuing to read requests it can no longer answer.

PR #227, opened 2026-05-26
Handle cross-language file diffs

### LINKED ISSUE #186: cross language diff misattributes orphan
## Summary

`sem diff a.ts b.py` (two-file no-git mode comparing files of different languages) attributes the deleted TS content to `entityId: "b.py::orphan"` even though that content never lived in `b.py`. There is also no warning that the two files are in different languages — which is almost certainly a user error worth flagging.

## Severity

P2 — wrong attribution (orphan IDed to a file whose parser never saw that source) + missing diagnostic for an almost-always-erroneous invocation.

## Environment

## Reproduction

```bash
repo=$(mktemp -d)
cd "$repo"

cat > a.ts <<'EOF'
function foo(x: number) { return x + 1; }
EOF

cat > b.py <<'EOF'
def foo(x): return x + 1
EOF

sem diff a.ts b.py --format json
```

## Observed

```json
{
  "changes": [
    {
      "entityId": "b.py::orphan",
      "filePath": "b.py",
      "beforeContent": "function foo(x: number) { return x + 1; }",
      "changeType": "deleted"
    }
  ]
}
```

The TS content lives under a `b.py::orphan` ID — sticking TypeScript source into a Python file path is misleading at best.

## Expected

Either:

1. **Reject mismatched languages** with a clear error: `error: cannot diff a.ts (TypeScript) against b.py (Python) — use --force to compare anyway`.
2. **Render as two separate sides** — show TS-side as deleted under the `a.ts` namespace, Python-side as added under `b.py`. The orphan synthesis under the *target* file is the bug.

Option 2 is more useful when users genuinely want a cross-language comparison (e.g., porting a function); option 1 is the safer default.

## Code references

The two-file file-compare branch is in [crates/sem-cli/src/commands/diff.rs](https://github.com/Ataraxy-Labs/sem/blob/2447cf74f7ac35466ac613b5a7401b1bf87431de/crates/sem-cli/src/commands/diff.rs) — search for `FileCompare` / two-file pathway. The orphan synthesis happens when content cannot be parsed by the target plugin; the fix is to retain the source-side file path and language plugin instead of forcing both into the target's namespace.

## Notes

A related cleaner case: comparing two files of the *same* language without git works fine. This bug only surfaces on cross-language comparisons.

Iron-Ham:
## Summary

- Detect language mismatches in two-file no-git diff mode and render them as source-side deletion plus target-side addition.
- Keep same-language two-file comparisons on the existing modified-target-file path.
- Normalize trailing `--format`/`--json` in `sem diff` args so the reported reproduction command works as written.

Closes #186

## Test plan

- `rustfmt --check crates/sem-cli/src/commands/diff.rs crates/sem-cli/tests/diff_file_compare.rs`
- `cargo test -p sem-cli`
- `cargo test --workspace`
- Created a dummy git repo under `~/Developer`, ran `sem diff a.ts b.py --format json`, and verified stderr warns about TypeScript/Python while JSON reports `a.ts::function::foo` deleted and `b.py::function::foo` added.

### DIFF
--- a/crates/sem-cli/src/commands/diff.rs
+++ b/crates/sem-cli/src/commands/diff.rs
@@ -6,9 +6,11 @@ use std::time::Instant;
 
 use sem_core::git::bridge::GitBridge;
 use sem_core::git::jj::maybe_resolve_ref;
-use sem_core::git::types::{DiffScope, FileChange};
+use sem_core::git::types::{DiffScope, FileChange, FileStatus};
 use sem_core::model::change::ChangeType;
 use sem_core::parser::differ::{compute_semantic_diff, DiffResult};
+use sem_core::parser::plugins::code::languages::get_language_config;
+use sem_core::parser::registry::{detect_ext_from_content, ParserRegistry};
 
 use crate::formatters::{
     json::format_json, markdown::format_markdown, plain::format_plain, terminal::format_terminal,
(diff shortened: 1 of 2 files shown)


Iron-Ham:
Reviewed the automated inspect notes. The split delete/add behavior for mismatched languages is intentional for #186: it preserves each side's file namespace instead of manufacturing a target-side orphan. The trailing `--format` handling also intentionally exits on missing or invalid values, matching normal CLI flag behavior; there is no later parser state to preserve after that error path.

Iron-Ham:
Addressed the inspect feedback around trailing `--format` handling by switching the normalizer to index-based parsing and adding a regression test for `--format --json`. The remaining latest inspect notes are on `process::exit(1)` error paths after the value check; there is no subsequent parser state because the process terminates. The cross-language delete/add status remains intentional for #186 so each side keeps its own file namespace.

rs545837:
Merged manually after rebasing onto main. Thanks!

PR #235, opened 2026-05-26
Include line spans in diff JSON

### LINKED ISSUE #188: diff json missing line numbers
## Summary

`sem diff --format json` does not include `startLine` / `endLine` on any change record. Consumers building patch summaries, PR comments, "click to navigate" links, or any UI keyed on file locations have no way to know *where* in the file each change occurred without separately calling `sem entities --json` and cross-referencing by `name + type` (which is not unique — see the JSON-schema-fragmentation issue).

## Severity

P2 — leaves consumers without basic location data. Workaround exists (call entities, join by name+type) but is fragile.

## Environment

## Reproduction

```bash
repo=$(mktemp -d)
cd "$repo"
git init -q && git config user.email t@t.com && git config user.name test

cat > a.ts <<'EOF'
function foo() { return 1; }
function bar() { return 2; }
EOF
git add -A && git commit -q -m v1

sed -i '' 's/return 1;/return 999;/' a.ts

sem diff --format json | jq '.changes[0] | keys'
```

## Observed

```
$ sem diff --format json | jq '.changes[0] | keys'
[
  "afterContent",
  "author",
  "beforeContent",
  "changeType",
  "commitSha",
  "entityId",
  "entityName",
  "entityType",
  "filePath",
  "oldEntityName",
  "oldFilePath",
  "oldParentId",
  "structuralChange"
]
```

No `startLine`, no `endLine`. `sem graph --json` and `sem entities --json` both include these fields; `sem diff --format json` does not.

## Expected

Each change record should carry:

```json
{
  "entityId": "a.ts::function::foo",
  "startLine": 1,
  "endLine": 1,
  "oldStartLine": 1,        // for renames / moves
  "oldEndLine": 1,
  ...
}
```

## Code references

[crates/sem-cli/src/formatters/json.rs](https://github.com/Ataraxy-Labs/sem/blob/2447cf74f7ac35466ac613b5a7401b1bf87431de/crates/sem-cli/src/formatters/json.rs) — the change-record serialization. The underlying `SemanticChange` (or equivalent) struct in sem-core does carry the entity, which has `start_line`/`end_line` — they're just dropped during JSON projection.

## Related

Part of the broader JSON-schema-fragmentation P1.

## Related Recent Fixes

Related to [#122](https://github.com/Ataraxy-Labs/sem/issues/122), which tracked normalization of JSON CLI contracts. On `sem 0.6.1`, `sem diff --format json` still omits `startLine` / `endLine` from change records, so this is a remaining schema-completeness issue after that broader JSON cleanup.

Iron-Ham:
## Summary

- Add current and previous line spans to semantic change records and expose them from `sem diff --format json` as `startLine`, `endLine`, `oldStartLine`, and `oldEndLine`.
- Preserve old spans for modified, moved, renamed, deleted, and reordered entities.
- Split orphan changes into aligned contiguous segments so reported line spans stay navigable.
- Update JSON output examples in README and docs.

Closes #188

## Test plan

- `cargo test --workspace`
- Built `sem-cli` and verified the issue #188 TypeScript reproduction in a dummy git repo under `/Users/zer0/Developer/oss`.
- Verified inserted orphan segment alignment in a dummy git repo under `/Users/zer0/Developer/oss`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,10 @@ sem diff --format json
       "changeType": "added",
       "entityType": "function",
       "entityName": "validateToken",
+      "startLine": 12,
+      "endLine": 18,
+      "oldStartLine": null,
+      "oldEndLine": null,
       "filePath": "src/auth.ts"
     }
   ]
--- a/crates/sem-cli/src/formatters/json.rs
+++ b/crates/sem-cli/src/formatters/json.rs
@@ -11,6 +11,10 @@ pub fn format_json(result: &DiffResult) -> String {
                 "changeType": c.change_type,
                 "entityType": c.entity_type,
                 "entityName": c.entity_name,
+                "startLine": c.start_line,
+                "endLine": c.end_line,
+                "oldStartLine": c.old_start_line,
+                "oldEndLine": c.old_end_line,
                 "oldEntityName": c.old_entity_name,
                 "filePath": c.file_path,
                 "oldFilePath": c.old_file_path,
@@ -46,6 +50,7 @@ pub fn format_json(result: &DiffResult) -> String {
 mod tests {
     use super::*;
     use sem_core::git::types::{FileChange, FileStatus};
+    use sem_core::model::change::SemanticChange;
     use sem_core::parser::differ::compute_semantic_diff;
     use sem_core::parser::plugins::create_default_registry;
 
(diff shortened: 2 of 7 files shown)


rs545837:
interesting thing to add, this is quite important on the ide integration view, and when human gets in the loop.

rs545837:
Squash-merged manually after rebasing onto latest main. All 264 tests pass.

PR #276, opened 2026-06-01
Classify same-name fuzzy matches as modified

### LINKED ISSUE #247: classify_match reports "renamed" for entities whose name never changed
**Severity:** High
**Component:** parser / entity identity (fuzzy match classification)
**Affected commands:** `sem diff`, `sem log`

## Summary

When the fuzzy-match phase re-pairs a before/after entity whose ID shifted for a non-rename reason, it classifies the pair as `Renamed` without ever checking that the name actually changed. The result is a bogus "renamed" change (with `oldEntityName` null) instead of a "modified", which misreports edits to a user or agent reading the diff.

## Reproduction

```sh
cat > over.ts <<'EOF'
function f(a: number): void {
  const x = a + 1;
  return process(x);
}
function f(a: string): void {
  const y = a.length;
  return handle(y);
}
EOF

# prepend one line (shifts both entities) and edit both bodies
cat > over.ts <<'EOF'
// header comment added
function f(a: number): void {
  const x = a + 100;
  return process(x);
}
function f(a: string): void {
  const y = a.length * 2;
  return handle(y);
}
EOF
sem diff over.ts --json
```

## Observed

The diff reports a `renamed` change for `f` with `oldEntityName` null (no name change), plus a spurious `deleted f` and `added f`. Summary: `modified=1, renamed=1, deleted=1, added=2`.

## Expected

Two same-name overloads whose bodies changed should be reported as modifications. With the identity-collision fix this case is `modified=2`; regardless, an entity whose name is unchanged must never be classified `Renamed`.

## Root cause

`classify_match` in `crates/sem-core/src/model/identity.rs:45-53` returns `ChangeType::Moved` when the file path or parent differs, and otherwise unconditionally returns `ChangeType::Renamed` — it never compares `before.name` to `after.name`:

```rust
fn classify_match(before: &SemanticEntity, after: &SemanticEntity) -> ChangeType {
    if before.file_path != after.file_path {
        ChangeType::Moved
    } else if before.parent_id != after.parent_id {
        ChangeType::Moved
    } else {
        ChangeType::Renamed
    }
}
```

It is reached from the fuzzy-match phase (around `crates/sem-core/src/model/identity.rs:253`, `:288`, `:386`). When an entity's ID changes for a non-rename reason — here the `@L{start_line}` suffix shifting because a line was prepended — exact and structural matching fail, fuzzy match re-pairs the entity, and `classify_match` tags it `Renamed` with no name change.

## Suggested fix

In `classify_match`, return `ChangeType::Modified` when `before.name == after.name` (same file and parent); only return `ChangeType::Renamed` when the names actually differ.

## Related issues

Related to #49 (closed, "Children symbols show as renamed even when not changed") and #20 (closed, rename detected as delete+add); the behavior reproduces on HEAD, so the earlier fix is incomplete for this path.

Iron-Ham:
Fixes #247

## Summary
- classify same-file, same-parent, same-name fuzzy matches as modified instead of renamed
- keep same-file, same-parent, different-name fuzzy matches classified as renamed
- add regression tests for both sides of that classifier boundary

## Test plan
- cargo test -p sem-core
- cargo test -p sem-cli
- git diff --check
- dummy git repo: issue #247 overload reproduction reports renamed: 0
- dummy git repo: true function rename still reports renamed: 1
- dummy git repo: renamed file with edited function still reports moved: 1
- dummy git repo: sem log for issue-style history reports added then modified (logic), no renamed entry

### DIFF
--- a/crates/sem-core/src/model/identity.rs
+++ b/crates/sem-core/src/model/identity.rs
@@ -47,6 +47,8 @@ fn classify_match(before: &SemanticEntity, after: &SemanticEntity) -> ChangeType
         ChangeType::Moved
     } else if before.parent_id != after.parent_id {
         ChangeType::Moved // intra-file scope move (e.g. method moved between classes)
+    } else if before.name == after.name {
+        ChangeType::Modified
     } else {
         ChangeType::Renamed
     }
@@ -143,9 +145,9 @@ fn make_change(
 
 /// Entity matching algorithm:
 /// 1. Exact ID match — same entity ID in before/after → modified or unchanged
-/// 2. Content hash match — same hash, different ID → renamed or moved
+/// 2. Content hash match — same hash, different ID → modified, renamed, or moved
 /// 3. Same signature across file rename → moved, even if content changed
-/// 4. Fuzzy similarity — >80% content similarity → probable rename
+/// 4. Fuzzy similarity — >80% content similarity → modified, renamed, or moved
 pub fn match_entities(
     before: &[SemanticEntity],
     after: &[SemanticEntity],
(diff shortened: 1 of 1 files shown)


Iron-Ham:
Superseded by #312.

PR #278, opened 2026-06-01
Move SQLite cache outside working trees

### LINKED ISSUE #251: Cache directory .sem is written into the user's working tree and is not gitignored
**Severity:** Medium
**Component:** cache / on-disk cache location
**Affected commands:** `sem graph`, `sem impact`, `sem context`, `sem verify`, `sem mcp` (any cache-building command)

## Summary

Cache-building commands write `.sem/cache.db` (a SQLite file that grows with repository size) into the root of the user's working tree as an untracked, non-ignored file. `sem` never adds `.sem/` to the user's `.gitignore`, so a routine `git add -A && git commit` will commit the cache. This contradicts the "works in any Git repo with no setup" positioning and is an easy footgun for users and agents.

## Reproduction

```sh
printf 'def f(a, b):\n    return a + b\n' > a.py

git status --porcelain          # clean
sem graph >/dev/null
git status --porcelain          # now shows ?? .sem/
git check-ignore .sem/cache.db && echo "ignored" || echo "NOT ignored"
```

## Observed

Before the cache build the tree is clean. After `sem graph`, `git status --porcelain` shows `?? .sem/` and `git check-ignore .sem/cache.db` reports it is not ignored. The `sem` repository ignores `.sem/` in its own `.gitignore`, so maintainers never see this in their own tree — but a user's repo gains an untracked, committable binary.

## Expected

The cache must not become an untracked, committable artifact in the user's working tree with zero setup.

## Root cause

`DiskCache::open` (`crates/sem-cli/src/cache.rs:23-27`) computes the cache directory as `repo_root.join(".sem")`, calls `std::fs::create_dir_all` on it, and opens `cache.db` inside; the MCP server's `DiskCache::open` does the same at `crates/sem-mcp/src/cache.rs:220-223`. The cache is therefore always created inside the user's repository root. `sem` never appends `.sem/` to the repository's `.gitignore`, so the untracked SQLite file (which grows with repository size) is staged by a routine `git add -A` and committed. `sem`'s own repository `.gitignore` lists `.sem/`, so maintainers never observe this in their own tree even though every user's repository is affected.

## Suggested fix

Either append `.sem/` to the repository's `.gitignore` on first cache write, or write the cache outside the repository entirely — for example into a per-repo directory under the OS cache directory, keyed by the repo's canonical path.

## Related issues

A separate issue covers the related case where a repo-controlled `.sem` symlink redirects cache writes; both stem from writing and resolving `.sem` inside the user's repository, and relocating the cache outside the repo resolves both.

Iron-Ham:
## Summary
- move CLI and MCP SQLite graph caches out of repo-local .sem directories into per-repo external cache roots
- ignore cache-root candidates that resolve inside the target repository, including repo-local SEM_CACHE_DIR overrides
- add regression coverage for clean git status after cache-building commands and document SEM_CACHE_DIR

Closes #251

## Test plan
- rustfmt --check sem-mcp/src/cache.rs sem-cli/src/cache.rs sem-cli/tests/cache_location.rs
- cargo test -p sem-mcp cache::tests
- cargo test -p sem-cli cache::tests
- cargo test -p sem-cli --test cache_location
- manual dummy repos: graph/impact/context/verify with external SEM_CACHE_DIR and graph with SEM_CACHE_DIR=.sem, verifying clean git status and no repo-local .sem

## Notes
- cargo clippy -p sem-mcp -p sem-cli --all-targets -- -D warnings currently fails on existing sem-core lints unrelated to this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ If you installed via npm/bun, the binary lives in `node_modules/.bin/sem` and is
 
 Works in any Git repo. No setup required. Also works outside Git for arbitrary file comparison.
 
+sem stores its SQLite entity cache outside the repository, under the OS cache directory by default. Set `SEM_CACHE_DIR=/path/to/cache` to override the cache root; repo-local overrides are ignored so cache files do not dirty the working tree.
+
 ### sem diff
 
 Entity-level diff with rename detection, structural hashing, and word-level inline highlights.
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -21,8 +21,10 @@ pub struct DiskCache {
 
 impl DiskCache {
     pub fn open(repo_root: &Path) -> Result<Self, rusqlite::Error> {
-        let cache_dir = repo_root.join(".sem");
-        std::fs::create_dir_all(&cache_dir).ok();
+        let cache_dir = shared_cache::cache_dir_for_repo(repo_root)
+            .ok_or_else(|| rusqlite::Error::InvalidPath(repo_root.to_path_buf()))?;
+        std::fs::create_dir_all(&cache_dir)
+            .map_err(|_| rusqlite::Error::InvalidPath(cache_dir.clone()))?;
         let db_path = cache_dir.join("cache.db");
         let conn = Connection::open(db_path)?;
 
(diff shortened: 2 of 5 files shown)


Iron-Ham:

PR #279, opened 2026-06-01
Handle Swift interpolated braces during recovery

### LINKED ISSUE #259: Swift conditional-compilation recovery drops the enclosing type when an interpolated string contains a brace
**Severity:** Medium
**Component:** parser / Swift entity extraction (conditional-compilation recovery)
**Affected commands:** `sem entities`, `sem graph`, `sem impact`, `sem context`, `sem diff`

## Summary

A Swift type whose body contains both a `#if/#endif` directive and a string-interpolation segment that includes a brace, such as `"\("}")"`, is dropped entirely from the entity tree: the type vanishes and its members are reparented to file scope. The trigger is narrow, but the blast radius is the whole enclosing type and all its members, and `#if` is common in cross-platform Swift.

## Reproduction

```sh
cat > Brace.swift <<'EOF'
class Config {
    let tpl = "prefix \("}") suffix"
#if DEBUG
    func dump() { print(tpl) }
#endif
    func render() -> String { return tpl }
}

struct Tail { let q: Int }
EOF
sem entities Brace.swift
```

Two controls confirm the isolation. The identical class without `#if` parses correctly:

```sh
cat > NoCond.swift <<'EOF'
class Config {
    let tpl = "prefix \("}") suffix"
    func render() -> String { return tpl }
}
EOF
sem entities NoCond.swift
```

The identical class with `#if` but a brace-free interpolated string also recovers correctly:

```sh
cat > Interp.swift <<'EOF'
class Greeter {
    let label = "outer \("inner") tail"
#if DEBUG
    func shout() { print(label) }
#endif
    func greet() -> String { return label }
}

struct After { let z: Int }
EOF
sem entities Interp.swift
```

## Observed

In the failing case, `class Config` has vanished and its three members are orphaned to file scope.

```
property  tpl     parent=<none>     ← orphaned
function  dump    parent=<none>     ← orphaned
function  render  parent=<none>     ← orphaned
struct    Tail    ...               ← unaffected
```

The `NoCond` control yields `Config` with all members nested; the `Interp` control yields `Greeter` with all members nested. So the defect is isolated to the recovery brace matcher when an interpolation containing a brace is present.

## Expected

`class Config` should be emitted with `tpl`, `dump`, and `render` as its children, exactly as in the two control cases.

## Root cause

tree-sitter-swift fails to keep the type intact when its body contains `#if/#else/#endif`, so `sem` has a hand-rolled recovery pass `recover_swift_conditional_compilation_containers` (`crates/sem-core/src/parser/plugins/code/entity_extractor.rs:529`) that re-parses the container by brace-matching the source. The matcher `find_matching_brace` (`entity_extractor.rs:768-814`) correctly skips line/block comments and multiline/single-line strings, but the single-line string skipper `skip_swift_string` (`entity_extractor.rs:860-871`) treats `\(` as an ordinary 2-character escape (`entity_extractor.rs:862-863`). For `"\("}")"` it therefore consumes `\(` as an escape, then reads the interpolation's opening `"` as the closing quote of the outer string and returns early. The `}` that is actually inside the interpolation is then seen as code and decrements the brace depth prematurely. `find_matching_brace` returns a truncated span, so `swift_contains_conditional_directive` no longer sees the `#if` in `parse_swift_recovered_container` (`entity_extractor.rs:644-648`), recovery returns `None`, the container is never recovered, and the orphaned members keep `parent_id: None`.

## Suggested fix

Teach `skip_swift_string` about interpolation: on encountering `\(` inside a string, recurse with paren-depth tracking (skipping nested strings) until the matching `)`, then resume string scanning. Optionally handle `#"…"#` raw strings, which are a related unhandled case in the same matcher (the mirror input `"\("{")"` would over-count and prevent the matcher from ever closing).

## Related issues

Edge case left unhandled by #230 ("Recover Swift containers with conditional compilation", which closed #192) — the recovery brace-matcher mishandles string interpolation `\("}")` and raw strings.

Iron-Ham:
## Summary
- Teach Swift conditional-compilation container recovery to skip full string literals, including interpolation and raw string delimiters
- Cap nested interpolation scanning to avoid pathological recursion
- Add regression coverage for regular, raw, multiline, and closure interpolation brace cases

Closes #259

## Test plan
- `cargo test -p sem-core test_swift_conditional_compilation_with_interpolated_brace_string -- --nocapture`
- `cargo test -p sem-core test_swift_conditional_compilation_inside_struct -- --nocapture`
- `cargo check -p sem-core`
- `cargo test -p sem-core`
- `cargo build -p sem-cli`
- Dummy git repos under `/tmp/sem-issue259-validation`: verified `sem entities --format json` parent IDs for regular, no-conditional, plain interpolation, raw interpolation, multiline interpolation, and closure interpolation cases
- Dummy git repo downstream checks: `sem graph --json --no-cache`, `sem impact render --file Brace.swift --json --no-cache`, `sem context Config --file Brace.swift --json --no-cache`, and `sem diff --json` after editing `Config.render`

### DIFF
--- a/crates/sem-core/src/parser/plugins/code/entity_extractor.rs
+++ b/crates/sem-core/src/parser/plugins/code/entity_extractor.rs
@@ -783,13 +783,8 @@ fn find_matching_brace(source: &[u8], open_brace: usize) -> Option<usize> {
             continue;
         }
 
-        if is_swift_multiline_string_start(source, i) {
-            i = skip_swift_multiline_string(source, i + 3);
-            continue;
-        }
-
-        if source[i] == b'"' {
-            i = skip_swift_string(source, i + 1);
+        if let Some(string_start) = swift_string_literal_start(source, i) {
+            i = skip_swift_string_literal(source, string_start);
             continue;
         }
 
@@ -799,7 +794,7 @@ fn find_matching_brace(source: &[u8], open_brace: usize) -> Option<usize> {
                 i += 1;
             }
             b'}' => {
-                depth -= 1;
+                depth = depth.checked_sub(1)?;
                 i += 1;
                 if depth == 0 {
                     // Return the byte after the closing brace for use as a slice end.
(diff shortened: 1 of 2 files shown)


Iron-Ham:
Superseded by #310.

PR #280, opened 2026-06-01
Normalize empty diff JSON output

### LINKED ISSUE #274: `sem diff --json` emits two different JSON shapes for an empty result
**Severity:** Low
**Component:** sem diff (JSON output)
**Affected commands:** `sem diff --json`

## Summary

`sem diff --json` produces two different key orderings for an empty result depending on how that empty result is reached. When the working tree is clean it prints a hardcoded JSON string with `summary` before `changes` and the summary keys in semantic order; when the result becomes empty after filtering (for example via `--no-cosmetics`) it goes through serde and prints `changes` before `summary` with the summary keys sorted alphabetically. An agent or golden test that diffs this output will see spurious changes.

## Reproduction

```sh

sem diff --json                                  # empty via clean worktree (hardcoded string)
printf 'def foo():\n        return 1\n' > app.py  # whitespace-only change
sem diff --json --no-cosmetics                   # empty via the serde path
```

## Observed

```text
clean worktree:   {"summary":{"fileCount":0,"added":0,"modified":0,"deleted":0,"moved":0,"renamed":0,"reordered":0,"orphan":0,"total":0},"changes":[]}
--no-cosmetics:   {"changes":[],"summary":{"added":0,"deleted":0,"fileCount":0,"modified":0,"moved":0,"orphan":0,"renamed":0,"reordered":0,"total":0}}
```

Two different orderings — both top-level key order (`summary`/`changes` vs `changes`/`summary`) and summary key order (semantic vs alphabetical) — for the same logical "no changes" result.

## Expected

A single, stable JSON shape for an empty result regardless of how it is produced, so the output is safe to diff, cache, and assert against.

## Root cause

The early empty-result path prints a hardcoded JSON string (`crates/sem-cli/src/commands/diff.rs:1029`) with `summary` first and the summary keys in semantic order. The normal path serializes the result through serde via `format_json` (`crates/sem-cli/src/commands/diff.rs:1056`), which emits `changes` first and sorts the summary keys alphabetically. The two paths never share a serializer.

## Suggested fix

Route the empty case through the same serializer as the non-empty case, for example `format_json(&DiffResult::default())`, and remove the hardcoded JSON string so there is exactly one shape.

## Related issues

Related to #85 (closed, "no changes output not JSON") and #122 (closed, "Normalize JSON CLI contracts"). The empty-result shape inconsistency above was not covered.

Iron-Ham:
## Summary

- route the no-file-change `sem diff --json` fast path through the shared JSON formatter
- add a regression test comparing clean-repo JSON with a `--no-cosmetics` filtered-empty result

Closes #274

## Test plan

- `cargo test -p sem-cli --test diff_patch json_empty_result_shape_matches_when_clean_or_filtered_empty`
- `cargo test -p sem-core format::json`
- `cargo test -p sem-cli --bin sem formatters::json`
- `cargo test -p sem-cli`
- Manual disposable git repos: compared clean, filtered-empty, pathspec-empty, and second clean-repo `sem diff --json` output byte-for-byte

### DIFF
--- a/crates/sem-cli/src/commands/diff.rs
+++ b/crates/sem-cli/src/commands/diff.rs
@@ -1026,7 +1026,7 @@ fn run_diff_pipeline(
     if file_changes.is_empty() {
         match opts.format {
             OutputFormat::Json => {
-                println!("{{\"summary\":{{\"fileCount\":0,\"added\":0,\"modified\":0,\"deleted\":0,\"moved\":0,\"renamed\":0,\"reordered\":0,\"orphan\":0,\"total\":0}},\"changes\":[]}}");
+                println!("{}", format_json(&DiffResult::default()));
             }
             _ => {
                 println!("\x1b[2mNo semantic changes detected.\x1b[0m");
--- a/crates/sem-cli/tests/diff_patch.rs
+++ b/crates/sem-cli/tests/diff_patch.rs
@@ -137,6 +137,24 @@ fn patch_mode_rejects_truncated_git_binary_patch() {
 }
 
 #[test]
+fn json_empty_result_shape_matches_when_clean_or_filtered_empty() {
+    let repo = TempRepo::new();
+    std::fs::write(repo.path.join("app.py"), "def foo():\n    return 1\n").expect("write app");
+    run_git(&repo.path, &["add", "app.py"]);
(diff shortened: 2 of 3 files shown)


Iron-Ham:

PR #281, opened 2026-06-01
Align patch pathspec matching with Git

### LINKED ISSUE #269: `sem diff --patch` pathspec matching diverges from git (directory over-match, no bracket classes)
**Severity:** Medium
**Component:** sem diff --patch (pathspec filtering)
**Affected commands:** `sem diff --patch -- <pathspec>`

## Summary

The pathspec filter used by `sem diff --patch` does not match git semantics in two ways: a trailing-slash directory spec like `src/` matches sibling directories such as `src2/` because the comparison is a bare `starts_with`, and the hand-rolled glob does not implement `[...]` character classes, so a pathspec like `a[12].py` that matches under the git-backed code path matches nothing under `--patch`. The two pathspec engines in the same tool disagree.

### Finding A — trailing-slash directory spec over-matches a sibling directory

```sh
mkdir -p src src2
printf 'def a():\n    return 1\n' > src/a.py
printf 'def b():\n    return 1\n' > src2/a.py
printf 'def a():\n    return 2\n' > src/a.py
printf 'def b():\n    return 2\n' > src2/a.py

git diff --name-only -- src/                          # git: only src/a.py
git diff | sem diff --patch --json -- src/            # sem: src/a.py AND src2/a.py
```

### Finding B — bracket character classes are ignored

```sh
printf 'def f():\n    return 1\n' > a1.py
printf 'def g():\n    return 1\n' > a2.py
printf 'def f():\n    return 2\n' > a1.py && git add a1.py
printf 'def g():\n    return 2\n' > a2.py && git add a2.py

sem diff --staged --json -- 'a[12].py'                 # git-backed path: matches a1.py + a2.py
git diff --cached | sem diff --patch --json -- 'a[12].py'  # --patch glob: matches nothing
```

### Finding A

```text
git:  src/a.py
sem:  ['src/a.py', 'src2/a.py']
```

### Finding B

```text
git-backed pathspec a[12].py: ['a1.py', 'a2.py']
--patch (sem glob)  a[12].py: []
```

## Expected

`src/` should match only files under the `src/` directory, as git does. `a[12].py` should match `a1.py` and `a2.py` under both code paths, matching git's bracket-class behavior; at minimum the two pathspec engines should agree.

## Root cause

Both defects live in `crates/sem-cli/src/commands/diff.rs`:

- **Finding A:** `path_matches_spec` (`crates/sem-cli/src/commands/diff.rs:133-141`) handles a trailing-slash directory spec with `file_path.starts_with(spec.trim_end_matches('/'))` at line 137. Because `"src2/a.py".starts_with("src")` is true, the spec leaks across the directory boundary. There is no component-boundary check.
- **Finding B:** `path_matches_spec` routes any spec containing `[` to `glob_match` (line 134), but `glob_match_inner` (`crates/sem-cli/src/commands/diff.rs:89-130`) only implements `*`, `**`, and `?`; a `[` falls through to the literal-character branch (lines 128-129), so `a[12].py` is matched as the literal text `a[12].py` and matches nothing.

## Suggested fix

- For directory specs, require a path-component boundary: match `spec == file_path` or `file_path.starts_with(format!("{spec}/"))` after trimming the trailing slash (the same boundary the non-glob branch already uses for plain specs).
- Implement `[...]` character classes in `glob_match_inner` (including ranges and negation), or document that `--patch` pathspecs do not support bracket classes and reject them rather than silently matching nothing.

## Related issues

Related to #136 (closed, "sem diff --patch ignores positional pathspec filters") fixed by #163 (merged) — the pathspec filtering added there diverges from git in the two ways above.

Iron-Ham:
## Summary

- Fix `sem diff --patch` directory pathspec filtering so `src/` does not match sibling prefixes like `src2/`.
- Add bracket-class pathspec matching, including ranges, negation, POSIX classes, and exact literal paths containing brackets.
- Align patch-mode wildcard matching with Git default pathspec behavior for nested paths.

Closes #269

## Test plan

- `cargo test -p sem-cli glob_matching_supports_bracket_classes && cargo test -p sem-cli pathspec_matching_respects_directory_boundaries`
- `cargo test -p sem-cli`
- Manual temp-repo parity checks comparing `git diff --name-only -- <pathspec>` with `git diff | sem diff --patch --json -- <pathspec>` for `src/`, `*.py`, `a[12].py`, and `a[[:digit:]].py`

### DIFF
--- a/crates/sem-cli/src/commands/diff.rs
+++ b/crates/sem-cli/src/commands/diff.rs
@@ -78,12 +78,84 @@ fn split_on_separator(args: Vec<String>) -> (Vec<String>, Vec<String>) {
     }
 }
 
-/// Simple glob matching for pathspecs. Supports `*` (any chars except `/`),
-/// `**` (any chars including `/`), and `?` (one non-`/` char).
+/// Simple glob matching for pathspecs. Supports `*`, `**`, `?`, and bracket classes.
 fn glob_match(pattern: &str, path: &str) -> bool {
     let pat: Vec<char> = pattern.chars().collect();
     let text: Vec<char> = path.chars().collect();
     glob_match_inner(&pat, &text)
+}
+
+fn posix_bracket_class_matches(name: &str, ch: char) -> Option<bool> {
+    match name {
+        "alnum" => Some(ch.is_ascii_alphanumeric()),
+        "alpha" => Some(ch.is_ascii_alphabetic()),
+        "blank" => Some(ch == ' ' || ch == '\t'),
+        "cntrl" => Some(ch.is_ascii_control()),
+        "digit" => Some(ch.is_ascii_digit()),
+        "graph" => Some(ch.is_ascii_graphic()),
+        "lower" => Some(ch.is_ascii_lowercase()),
+        "print" => Some(ch.is_ascii_graphic() || ch == ' '),
+        "punct" => Some(ch.is_ascii_punctuation()),
+        "space" => Some(ch.is_ascii_whitespace()),
+        "upper" => Some(ch.is_ascii_uppercase()),
+        "xdigit" => Some(ch.is_ascii_hexdigit()),
+        _ => None,
+    }
+}
+
+fn match_posix_bracket_class(pat: &[char], ch: char) -> Option<(bool, usize)> {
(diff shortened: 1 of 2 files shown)


Iron-Ham:

PR #284, opened 2026-06-01
Parse diff ref path arguments

### LINKED ISSUE #270: `sem diff <ref> <path>` is parsed as a revision range, unlike `git diff`
**Severity:** Medium
**Component:** sem diff (positional argument parsing)
**Affected commands:** `sem diff <ref> <path>`, `sem diff <ambiguous-name>`

## Summary

`sem diff --help` advertises that it "supports git diff syntax", but `sem diff HEAD app.py` fails with a revision-not-found error where `git diff HEAD app.py` works, because two positional arguments are only treated as a ref-plus-path pair when both happen to be existing files on disk; otherwise both are parsed as a revision range. A related, lower-stakes case is the single-argument name that is both a valid ref and an on-disk file: git aborts as ambiguous, while sem resolves it to a pathspec.

### Finding A — ref plus path is parsed as a range

```sh
printf 'def foo():\n    return 2\n' > app.py

git diff HEAD app.py             # works
sem diff HEAD app.py --json      # fails
sem diff HEAD --json -- app.py   # works (the required workaround)
```

### Finding B — single ambiguous name

```sh
printf 'def foo():\n    return 2\n' > app.py && git commit -qam c2
git branch app.py HEAD~1         # a branch named exactly like the file
printf 'def foo():\n    return 3\n' > app.py

git diff app.py                  # git aborts: ambiguous argument
sem diff app.py --json           # sem silently treats app.py as a pathspec
```

### Finding A

```text
git diff HEAD app.py    -> normal diff, exit 0
sem diff HEAD app.py    -> Error: git error: revspec 'app.py' not found ...  (exit 1)
```

### Finding B

```text
git diff app.py  -> fatal: ambiguous argument 'app.py': both revision and filename  (exit 128)
sem diff app.py  -> diffs the working tree against HEAD for path app.py, exit 0
```

## Expected

`sem diff HEAD app.py` should behave like `git diff HEAD app.py`: treat `HEAD` as a revision and `app.py` as a path, without requiring an explicit `--` separator (Finding A). For the single ambiguous name (Finding B), sem's path-first resolution is a reasonable and predictable design choice (a real file usually beats a coincidentally-named branch); the only divergence is that git refuses to guess. Aligning with git's hard error is not necessary — a one-line stderr note when an argument is both a valid ref and an on-disk path would be sufficient.

## Root cause

In `parse_args` in `crates/sem-cli/src/commands/diff.rs`:

- **Finding A:** the two-argument branch (`crates/sem-cli/src/commands/diff.rs:303-328`) only produces a `FileCompare` when *both* positionals exist as files on disk (lines 308-322); otherwise it falls through to `Range(a, b)` (lines 325-328). `HEAD` is not a file, so `HEAD app.py` becomes a range and `app.py` is later handed to git as a revision, which fails. The actual error is raised downstream when the range scope is resolved (around `crates/sem-cli/src/commands/diff.rs:931`).
- **Finding B:** the single-argument branch checks `Path::new(cwd).join(arg).exists()` before any ref resolution (`crates/sem-cli/src/commands/diff.rs:277-284`), so an existing file always wins over a same-named ref, with no ambiguity detection.

## Suggested fix

- In the two-argument branch, disambiguate before defaulting to a range: if the first argument resolves as a revision and the second does not, treat the second as a pathspec (ref-plus-path), matching git. Alternatively, narrow the `--help` text to state that `--` is required to separate a path from a revision.
- For Finding B, optionally emit a stderr note when a single argument is both a valid ref and an on-disk path, to make the path-first choice explicit. No behavioral change required.

## Related issues

Related to #135 / #134 / #170 (closed, sem diff argument handling) fixed by #163 (merged), and to #36 / #37 (git-diff compatibility). None covered the ref-plus-path two-argument case or the ambiguous single-name case.

Iron-Ham:
Closes #270

## Summary
- Parse `sem diff <ref> <path>` as a ref-to-working-tree diff filtered by the pathspec when the first positional argument is a revision and the second is an existing path or explicit pathspec.
- Warn when a single positional argument is both an on-disk path and a Git revision, while continuing to treat it as a pathspec.
- Preserve missing-path errors and two-revision range parsing.

## Test plan
- `rustfmt --check crates/sem-cli/src/commands/diff.rs crates/sem-cli/tests/diff_file_compare.rs`
- `cargo test -p sem-cli --test diff_file_compare`
- `cargo test -p sem-cli`
- Manual dummy Git repos validating `sem diff HEAD app.py`, `sem diff HEAD missing.py`, `sem diff HEAD~1 HEAD`, and the branch/file ambiguity warning.

### DIFF
--- a/crates/sem-cli/src/commands/diff.rs
+++ b/crates/sem-cli/src/commands/diff.rs
@@ -78,6 +78,22 @@ fn split_on_separator(args: Vec<String>) -> (Vec<String>, Vec<String>) {
     }
 }
 
+fn looks_like_pathspec(arg: &str) -> bool {
+    arg.starts_with(':') || arg.contains('*') || arg.contains('?') || arg.contains('[')
+}
+
+fn get_git_bridge<'a>(git: &'a mut Option<Option<GitBridge>>, cwd: &str) -> Option<&'a GitBridge> {
+    if git.is_none() {
+        *git = Some(GitBridge::open(Path::new(cwd)).ok());
+    }
+
+    git.as_ref().and_then(|git| git.as_ref())
+}
+
+fn is_valid_git_rev(git: &mut Option<Option<GitBridge>>, cwd: &str, arg: &str) -> bool {
+    get_git_bridge(git, cwd).is_some_and(|git| git.is_valid_rev(arg))
+}
+
 /// Simple glob matching for pathspecs. Supports `*` (any chars except `/`),
 /// `**` (any chars including `/`), and `?` (one non-`/` char).
 fn glob_match(pattern: &str, path: &str) -> bool {
@@ -232,9 +248,9 @@ fn normalize_trailing_output_format(opts: &mut DiffOptions) {
     opts.args = args;
 }
 
-
 fn parse_args(args: Vec<String>, cwd: &str) -> ParsedArgs {
     let (refs, pathspecs) = split_on_separator(args);
+    let mut git = None;
 
     if refs.is_empty() {
         return ParsedArgs {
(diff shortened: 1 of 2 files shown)


Iron-Ham:

PR #285, opened 2026-06-01
Extract TypeScript namespace modules

### LINKED ISSUE #263: TypeScript `namespace` / `module` declarations are not extracted; members are silently flattened to top level
**Severity:** High
**Component:** parser / TypeScript entity extraction, graph
**Affected commands:** `sem entities`, `sem diff`, `sem impact`, `sem context`, `sem graph`

## Summary

`namespace X {}` parses as the tree-sitter node `internal_module`, and `module X {}` / `declare module "x" {}` parse as `module`. Neither is in the TypeScript/TSX `entity_node_types`, so the namespace/module entity is dropped. Because the body is a `statement_block` (which is a configured container type), the members are still walked and emitted as top-level entities with `parent_id: null`, and the namespace qualifier is lost. This is worse than erroring: results look plausible but are wrong. `Geometry.distance` is reported as a bare `distance`; two namespaces each containing a `helper` would both surface as a top-level `helper`, disambiguated only by line. Qualified calls like `Utils.parse()` are not resolved, and `sem impact` on the namespace name fails.

## Reproduction

```sh
cat > namespace.ts <<'EOF'
export namespace Utils {
  export function parse(input: string): string {
    return input.trim();
  }
  export class Parser {
    run(input: string): string { return parse(input); }
  }
}

export function useNamespace(input: string): string {
  return Utils.parse(input);
}

module LegacyModule {
  export function helper(): void {}
}
EOF

sem entities namespace.ts
sem impact Utils --file namespace.ts --no-cache
sem impact useNamespace --file namespace.ts --no-cache
sem graph --json --no-cache
```

## Observed

`sem entities namespace.ts`:

```
  function parse (L2:4)
  class Parser (L5:7)
    method run (L6:6)
  function useNamespace (L10:12)
  function helper (L15:15)
```

There is no `Utils` and no `LegacyModule` entity; `parse`, `useNamespace`, `Parser`, and `helper` are all top-level. `sem impact Utils` prints `error: Entity 'Utils' not found`. The graph contains `Parser -> parse` and `Parser::run -> parse` (the unqualified internal call resolves by name) but no `useNamespace -> parse` edge — the qualified `Utils.parse()` call is unresolved. `sem impact useNamespace` reports no dependencies.

## Expected

`Utils` and `LegacyModule` are emitted as `module`-typed entities, with `parse`, `Parser`, and `helper` nested under them with correct `parent_id`. `Utils.parse(input)` resolves to the namespace member, so `useNamespace -> parse` is an edge and `sem impact Utils` works.

## Root cause

`crates/sem-core/src/parser/plugins/code/languages.rs`: `internal_module` and `module` are absent from `TYPESCRIPT_CONFIG`/`TSX_CONFIG` `entity_node_types` (L412–426 and L439–453). In `crates/sem-core/src/parser/plugins/code/entity_extractor.rs`, `map_node_type` already maps `module` to `"module"` (L1568–1569) but `internal_module` is unmapped (would fall through to `other => other` at L1593). Because the configured `container_node_types` include `statement_block` (the body kind of both `internal_module` and `module`), the members are traversed and emitted but never re-parented, which is the flattening.

Qualified-call resolution is a second, related gap: in `crates/sem-core/src/parser/scope_resolve.rs`, the `alias.method()` path (L3250–3254) resolves through the import table for namespace imports, but local namespace declarations never populate an equivalent symbol table or scope, so `Utils.parse()` has nothing to resolve against.

The grammar (`tree-sitter-typescript` 0.23.2) confirms `internal_module` has a `name` field (`identifier` / `nested_identifier` / `string`) and a `statement_block` body. For `declare module "x"` the name is a string literal and needs a small extractor branch.

## Suggested fix

1. Add `"internal_module"` and `"module"` to `entity_node_types` in both configs; map `"internal_module" => "module"` in `map_node_type`.
2. Re-parent members under the namespace/module entity (the `statement_block` body is already a container).
3. Handle the `declare module "x"` string-literal name in `extract_name`.
4. Populate a local-namespace symbol table so `X.f()` resolves to the contained member; add tests for `namespace X { export function f() {} }` with `X.f()` from outside.

## Related issues

Distinct from #197 ("graph namespace imports false method edges") — that issue concerns namespace **imports**, whereas this is about local namespace/module **declarations** not being extracted at all.

Iron-Ham:
## Summary
- Extract TypeScript/TSX namespace and module declarations as module entities
- Preserve namespace/module parentage for nested declarations, including ambient module function signatures
- Resolve qualified namespace member calls such as `Utils.parse()`, including imported namespace entities

Fixes #263

## Test plan
- `cargo test --manifest-path crates/Cargo.toml -p sem-core`
- `cargo test --manifest-path crates/Cargo.toml -p sem-cli`
- Dummy git repo validation with `sem entities`, `sem graph --json --no-cache`, and `sem impact` for same-file namespace calls, imported namespace calls, and ambient `declare module` declarations


Iron-Ham:
Superseded by #309.

PR #287, opened 2026-06-01
Suppress Swift local property entities

### LINKED ISSUE #254: Swift local variables are extracted as `property` entities and corrupt the dependency graph
**Severity:** High
**Component:** parser / Swift entity extraction, graph
**Affected commands:** `sem entities`, `sem graph`, `sem impact`, `sem context`, `sem log`, `sem blame`, `sem diff`

## Summary

A `let`/`var` declared directly inside a Swift function or method body is emitted as a first-class entity of type `property`, parented to the enclosing function. This is semantically wrong (a Swift property is a member of a type, not a local binding), and because these phantom locals become real graph nodes the scope resolver builds false `calls` edges to and from them. Every Swift function that declares a local variable is affected, which is to say almost all real Swift code, so `sem impact`/`sem context` (the blast-radius and agent-context commands) return false dependents and addressable non-entities.

### Finding A — entity listing

```sh
cat > Example.swift <<'EOF'
struct Cache {
    let stored = 0
    func value() -> Int {
        let doubled = stored * 2
        let offset = doubled + 1
        return offset
    }
}
EOF
sem entities Example.swift
```

### Finding B — graph / impact / context corruption

```sh
cat > Mini.swift <<'EOF'
func connect() -> Int { return 0 }

func build() -> Int {
    let conn = connect()
    return conn
}
EOF
sem graph --no-cache --json --file-exts .swift
sem impact connect --no-cache --json
sem context conn
```

## Observed

Finding A: `doubled` and `offset` are listed as `property` entities parented to the function `value`.

```
  struct Cache (L1:7)
    property stored (L2:2)
    function value (L3:6)
      property doubled (L4:4)
      property offset (L5:5)
```

Finding B: the local `conn` becomes the node `Mini.swift::function::build::conn`, gets a false `calls` edge to `connect`, is reported as a dependent of `connect` by `sem impact`, and is addressable by `sem context`.

```
# graph edges
build       --calls--> connect
build::conn --calls--> connect      ← FALSE: a local variable "calls" connect

# impact connect
dependents: [ build, conn (property) ]   ← FALSE local-variable dependent

# context conn
target: property conn — let conn = connect()
```

## Expected

Local bindings are not type members. They should be folded into the containing function entity, not emitted as `property` entities, and must never become graph nodes, impact dependents, or `context`/`log`/`blame` targets.

## Root cause

In tree-sitter-swift a local `let x = …` and a stored property are both `property_declaration` nodes, differing only by parent. `SWIFT_CONFIG` is the only language config that lists `function_body` as a container (`crates/sem-core/src/parser/plugins/code/languages.rs:710`), declares `property_declaration` as an entity type (`languages.rs:706`), and leaves `suppressed_nested_entities` empty (`languages.rs:712`). The entity walker descends into every container child looking for nested entities (`crates/sem-core/src/parser/plugins/code/entity_extractor.rs:448-459`); because `function_body` is a container, it extracts the local `property_declaration` nodes. `function_body` is needed so nested functions still extract, so it should stay — only the local properties should be suppressed.

For comparison, Kotlin also models locals as `property_declaration` but its `container_node_types` is `["class_body", "enum_class_body"]` (`languages.rs:777`), with no function body, so Kotlin does not leak locals. Swift is the outlier.

## Suggested fix

Add suppression rules to `SWIFT_CONFIG.suppressed_nested_entities` keyed on the body-bearing parent node types, e.g. `SuppressedNestedEntity { parent_entity_node_type: "function_declaration", child_entity_node_type: "property_declaration" }`, and the same for `init_declaration`, `subscript_declaration`, and `deinit_declaration`. The suppression check (`entity_extractor.rs:1493-1502`) already keys on the parent entity's node type, which is exactly what is needed. Real properties (parent `class_declaration`/`struct_declaration`/etc.) are unaffected, and nested functions continue to extract.

## Related issues

Same defect class as #211 (C/C++ block-locals), fixed by #215 ("Suppress C and C++ block-local variables"); Swift was not covered there, and the same `suppressed_nested_entities` mechanism applies.

Iron-Ham:
Fixes #254

## Summary
- suppress Swift local `let`/`var` property declarations inside function-like bodies so they do not become semantic entities
- add stable extraction/scope names for Swift `deinit` and `subscript` declarations
- preserve nested declarations in Swift computed property, getter, and subscript bodies while suppressing local bindings
- consume scope-local bindings before graph fallback resolution to avoid false edges to same-named entities

## Test plan
- `cargo test -p sem-core swift -- --nocapture`
- `cargo test -p sem-core`
- dummy Swift git repos validated `sem entities`, `sem graph --no-cache --json --file-exts .swift`, and `sem impact ... --no-cache --json --file-exts .swift`

## Review
- independent code reviewer findings addressed
- `claude -p` reviewer was launched twice but produced no output and hung; both processes were terminated cleanly

### DIFF
--- a/crates/sem-core/src/parser/graph.rs
+++ b/crates/sem-core/src/parser/graph.rs
@@ -2304,6 +2304,59 @@ class UserService {
     }
 
     #[test]
+    fn test_swift_local_bindings_do_not_become_graph_nodes() {
+        let (dir, registry) = create_test_repo();
+        let root = dir.path();
+
+        write_file(root, "Mini.swift", "\
+func connect() -> Int { return 0 }
+func conn() -> Int { return 1 }
+
+func build() -> Int {
+    let conn = { 2 }
+    return conn() + connect()
+}
+");
+
+        let (graph, _) = EntityGraph::build(root, &["Mini.swift".into()], &registry);
+        let names: Vec<&str> = graph.entities.values().map(|entity| entity.name.as_str()).collect();
+
+        assert!(names.contains(&"connect"), "got: {:?}", names);
+        assert!(names.contains(&"conn"), "got: {:?}", names);
+        assert!(names.contains(&"build"), "got: {:?}", names);
+
+        let connect_id = graph
+            .entities
+            .values()
+            .find(|entity| entity.name == "connect")
+            .map(|entity| entity.id.clone())
+            .expect("connect entity should exist");
+        let dependents = graph.get_dependents(&connect_id);
+        let dependent_names: Vec<&str> = dependents.iter().map(|entity| entity.name.as_str()).collect();
+
+        assert!(dependent_names.contains(&"build"), "got: {:?}", dependent_names);
+        assert!(!dependent_names.contains(&"conn"), "got: {:?}", dependent_names);
+        assert!(
+            graph.edges.iter().all(|edge| !edge.from_entity.ends_with("::conn")),
+            "local binding should not create edges. Edges: {:?}",
+            graph.edges
+        );
+
+        let conn_id = graph
+            .entities
+            .values()
+            .find(|entity| entity.name == "conn")
+            .map(|entity| entity.id.clone())
+            .expect("conn entity should exist");
+        let conn_dependents = graph.get_dependents(&conn_id);
+        assert!(
+            conn_dependents.is_empty(),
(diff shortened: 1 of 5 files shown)


Iron-Ham:
Superseded by #310.

PR #288, opened 2026-06-01
Extract JavaScript object literal method entities

### LINKED ISSUE #266: Object-literal method shorthand is not extracted as nested entities
**Severity:** Low
**Affected commands:** `sem entities`, `sem diff` (granularity)

## Summary

For `const svc = { open(): void {}, close(): void {} }`, only the `variable svc` entity is produced; the `open` and `close` methods are not extracted as nested entities. This is inconsistent with class expressions assigned to a const (`const Widget = class { render() {} }`), where `render` is correctly nested under `Widget`. The common "export a service object of methods" pattern therefore loses per-method granularity: a change to `svc.open` shows up only as a change to the whole `svc` variable.

## Reproduction

```sh
cat > both.ts <<'EOF'
export const svc = {
  open(): void {},
  close(): void {},
};

export const Widget = class {
  render(): string { return "widget"; }
};
EOF

sem entities both.ts
```

## Observed

```
  variable svc (L1:4)
  variable Widget (L6:8)
    method render (L7:7)
```

The class-expression initializer (`Widget`) nests `render`, but the object-literal initializer (`svc`) yields no nested `open` / `close`.

## Expected

`open` and `close` are nested under `svc` (as methods), mirroring the class-expression path, so per-method diffs and entity lookups work.

## Root cause

`crates/sem-core/src/parser/plugins/code/entity_extractor.rs`: a variable declarator's initializer is descended into via `push_js_ts_initializer_children` / `js_ts_initializer_value` / `is_js_ts_initializer_node` (L1895–1944). `is_js_ts_initializer_node` (L1942–1944) returns true only when the value's kind is in `scope_boundary_types` or is exactly `"class"`. The JS/TS scope boundaries (`JS_TS_SCOPE_BOUNDARIES` in `crates/sem-core/src/parser/plugins/code/languages.rs:378-382`) are `arrow_function`, `function_expression`, and `generator_function` — `object` is not included. So function/arrow/generator/class initializers are descended into, but `object` initializers are not, and their `method_definition` children are never reached.

## Suggested fix

When a declarator's initializer is an `object`, traverse its `method_definition` children (and `pair`s whose value is a function or arrow) as nested entities under the variable, mirroring the class-expression path. Add a test asserting `const svc = { open() {}, close() {} }` nests `open` and `close` under `svc`.

## Related issues

Related to #232 / #3 (nested TypeScript class fields/methods) and #177 (anonymous nested classes flatten); object-literal initializers were not covered by that work.

Iron-Ham:
## Summary
- Extract JS/TS object-literal shorthand methods and function-valued properties under variable/field initializer entities
- Preserve nested diff granularity for object-literal methods while avoiding redundant variable container changes
- Tighten parent suppression to remove exact child content ranges so same-line parent/sibling changes remain visible

Fixes #266

## Test plan
- cargo test -p sem-core
- cargo test -p sem-cli
- Built target/debug/sem and validated sem entities / sem diff against disposable git repos for shorthand methods, arrow/function-valued object properties, ignored empty/computed keys, and inline parent/sibling changes

## Review
- Independent code_reviewer subagent reviewed the patch; addressed same-line suppression and missing pair-diff coverage findings
- claude -p reviewed the patch; addressed duplicate fallback and empty/dynamic object key hardening findings

### DIFF
--- a/crates/sem-core/src/parser/differ.rs
+++ b/crates/sem-core/src/parser/differ.rs
@@ -213,6 +213,7 @@ fn suppress_redundant_parents(
         "export",
         "package",
         "field",
+        "variable",
         "svelte_instance_script",
         "svelte_module_script",
         "object",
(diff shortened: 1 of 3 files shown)


Iron-Ham:
Superseded by #309.

PR #289, opened 2026-06-01
Distinguish JavaScript and TypeScript accessors

### LINKED ISSUE #267: TypeScript getter and setter accessors are indistinguishable (both emit as `method <name>`)
**Severity:** Low
**Affected commands:** `sem entities`, `sem context` / `sem impact` (by name)

## Summary

A property getter and setter both parse as `method_definition` and both emit as `method <name>` with the same name and type, differing only by line. `sem entities` shows two identical-looking `method value` rows, and `sem context value` / `sem impact value` on the name are ambiguous (which accessor?). `map_node_type` already has dedicated `getter` / `setter` entity types, but they only apply to Dart's `getter_signature` / `setter_signature`; TypeScript accessors never reach them. This is an identity/clarity defect, not a diff-correctness bug — `sem diff` still isolates a getter-only change because the entity content includes the `get value(...)` text.

## Reproduction

```sh
cat > acc.ts <<'EOF'
export class Box {
  private _v = 0;
  get value(): number { return this._v; }
  set value(n: number) { this._v = n; }
}
EOF

sem entities acc.ts
```

## Observed

```
  class Box (L1:5)
    field _v (L2:2)
    method value (L3:3)
    method value (L4:4)
```

Both the getter (L3) and the setter (L4) appear as `method value` with the same name and type; the `--json` form confirms both carry `"type": "method"` and the same parent. They are distinguishable only by line.

## Expected

The getter and setter are distinguishable — for example emitted as `getter value` and `setter value`, or with a `(get)` / `(set)` suffix on the name — so name-based lookups and context queries are unambiguous.

## Root cause

`crates/sem-core/src/parser/plugins/code/entity_extractor.rs`: TypeScript get/set accessors are `method_definition` nodes that carry a `get` / `set` keyword as an anonymous child (in the `tree-sitter-typescript` 0.23.2 grammar it is not a named field). `map_node_type` maps every `method_definition` to `"method"` (L1553) and never inspects that keyword. The `getter` / `setter` entity types exist in `map_node_type` (L1561–1562) but are only produced from Dart's `getter_signature` / `setter_signature`, so TypeScript accessors cannot reach them.

## Suggested fix

Detect the `get` / `set` keyword child on a TypeScript `method_definition` and assign `entity_type` `"getter"` / `"setter"` (the types already exist in `map_node_type`), or append `(get)` / `(set)` to the entity name. Add a test asserting that a class with both a `get value` and a `set value` produces two distinguishable entities.

## Related issues

Related to #153 (entity type vocabulary).

Iron-Ham:
Closes #267

## Summary
- classify JavaScript/TypeScript class accessors as getter/setter entities instead of generic methods
- allow context and impact lookups to use type-qualified entity queries such as `getter value` and `setter value`
- add parser and CLI regression coverage for TypeScript, TSX, and JavaScript accessors

## Test plan
- `cargo test -p sem-core --no-fail-fast`
- `cargo test -p sem-cli --no-fail-fast`
- validated the rebuilt `target/debug/sem` in temporary git repos for `entities`, `diff`, `context`, and `impact` accessor behavior

## Review
- independent subagent review completed; requested missing tests were added
- `claude -p` review was launched twice, but the CLI produced no output and timed out on the bounded retry

### DIFF
--- a/crates/sem-cli/src/commands/context.rs
+++ b/crates/sem-cli/src/commands/context.rs
@@ -120,7 +120,11 @@ fn find_entity<'a>(
         std::process::exit(1);
     });
 
-    let mut matching: Vec<_> = graph.entities.values().filter(|e| e.name == name).collect();
+    let mut matching: Vec<_> = graph
+        .entities
+        .values()
+        .filter(|e| super::entity_matches_query(e, name))
+        .collect();
 
     if matching.is_empty() {
         eprintln!("{} Entity '{}' not found", "error:".red().bold(), name);
@@ -146,7 +150,10 @@ fn find_entity<'a>(
     matching.sort_by_key(|e| (&e.file_path, e.start_line));
     eprintln!("{} Entity name '{}' is ambiguous ({} matches). Specify --file or --entity-id:", "error:".red().bold(), name, matching.len());
     for m in &matching {
-        eprintln!("  {} ({}:L{})", m.id, m.file_path, m.start_line);
+        eprintln!(
+            "  {} {} ({}:L{})",
+            m.entity_type, m.id, m.file_path, m.start_line
+        );
     }
     std::process::exit(1);
 }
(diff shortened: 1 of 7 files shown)


Iron-Ham:
Superseded by #309.

PR #290, opened 2026-06-01
Harden diff output escaping

### LINKED ISSUE #273: Untrusted source content is not escaped in `sem diff` markdown and terminal output
**Severity:** Low
**Component:** sem diff output formatters (`formatters/markdown.rs`, `formatters/terminal.rs`)
**Affected commands:** `sem diff --format markdown -v`, `sem diff -v` (terminal)

## Summary

`sem diff` embeds source content into its rendered output without escaping it. In verbose markdown, source containing a code-fence sequence closes the surrounding ```` ```diff ```` block early and corrupts the rendered document (for example when the output is pasted into a PR or issue). In terminal output, raw ANSI/control bytes from the source pass through unsanitized even with `--color never`, letting a hostile file manipulate the viewer's terminal. Both are output-hardening gaps rather than correctness bugs.

### Finding A — markdown fence break

```sh
printf 'def foo():\n    s = "plain"\n    return s\n' > app.py
fence='```'
printf 'def foo():\n    s = "X%s Y %s Z"\n    return s\n' "$fence" "$fence" > app.py

sem diff --format markdown -v
```

### Finding B — ANSI/control bytes leak with `--color never`

```sh
printf 'def foo():\n    s = "\033[31mRED\033[0m"\n    return s\n' > app.py

sem diff -v --color never | cat -v   # raw ESC[31m appears in the output
```

### Finding A

The embedded ```` ``` ```` sequences from the source appear verbatim inside the ```` ```diff ```` block, producing an odd number of fence lines and closing the fence prematurely, which corrupts the rendered markdown.

### Finding B

The raw `ESC[31m` escape sequence from the source is present in the output despite `--color never` (sem's own coloring is suppressed, but source bytes are not sanitized).

## Expected

Source content embedded in markdown should not be able to break out of its enclosing fence; the renderer should choose a fence longer than any backtick run in the content. Terminal output should strip or escape C0 control bytes (especially `\x1b`) from rendered source so a file cannot drive the viewer's terminal, particularly under `--color never`.

## Root cause

- **Finding A:** the verbose markdown formatter wraps content in a fixed-length ```` ```diff ```` fence with no escaping (`crates/sem-cli/src/formatters/markdown.rs:89-121`). Content containing an equal-or-longer backtick run terminates the fence early.
- **Finding B:** the terminal formatter emits rendered source content without stripping control bytes (`crates/sem-cli/src/formatters/terminal.rs:130-244`); `--color never` only governs sem's own color codes, not bytes originating in the source.

## Suggested fix

- In the markdown formatter, compute the longest run of backticks in the content and open/close the fence with at least one more backtick (the CommonMark rule), or escape embedded fences.
- In the terminal formatter, strip or visibly escape C0 control bytes (at minimum `\x1b`) from rendered content before printing.

## Related issues

Related to #68 / #51 (closed, markdown verbose output). Neither addressed fence escaping or terminal control-byte sanitization.

Iron-Ham:
## Summary
- use markdown code fences that grow beyond embedded source backtick runs
- escape source-derived terminal control characters before applying CLI colors
- cover source content and unsupported-file warning paths with dummy git repo regressions

Closes #273

## Test plan
- cargo test -p sem-cli --test diff_output_hardening
- cargo test -p sem-cli
- cargo test
- manual disposable git repos for markdown backticks and terminal ESC filename/content hardening

### DIFF
--- a/crates/sem-cli/src/formatters/markdown.rs
+++ b/crates/sem-cli/src/formatters/markdown.rs
@@ -4,6 +4,37 @@ use sem_core::parser::differ::DiffResult;
 use similar::{ChangeTag, TextDiff};
 use std::collections::BTreeMap;
 
+fn longest_backtick_run(input: &str) -> usize {
+    let mut longest = 0;
+    let mut current = 0;
+
+    for ch in input.chars() {
+        if ch == '`' {
+            current += 1;
+            longest = longest.max(current);
+        } else {
+            current = 0;
+        }
+    }
+
+    longest
+}
+
+fn push_diff_block(lines: &mut Vec<String>, diff_lines: Vec<String>) {
+    let fence_len = diff_lines
+        .iter()
+        .map(|line| longest_backtick_run(line))
+        .max()
+        .unwrap_or(0)
+        .saturating_add(1)
+        .max(3);
+    let fence = "`".repeat(fence_len);
+
+    lines.push(format!("{fence}diff"));
+    lines.extend(diff_lines);
+    lines.push(fence);
+}
+
 pub fn format_markdown(result: &DiffResult, verbose: bool) -> String {
     if result.changes.is_empty() {
         return "No semantic changes detected.".to_string();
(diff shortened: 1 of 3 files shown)


Iron-Ham:
Superseded by #311.

PR #291, opened 2026-06-01
Extract TypeScript ambient function signatures

### LINKED ISSUE #265: Bodyless TypeScript function signatures (overload heads and `declare function`) are dropped
**Severity:** Medium
**Affected commands:** `sem entities`, `sem diff`, `sem context`

## Summary

A function declaration with no body — an overload head (`function f(a: string): string;`) or an ambient declaration (`declare function f(x: number): number;`) — parses as the tree-sitter node `function_signature`, which is not in the TypeScript `entity_node_types`. Only the overload implementation (`function_declaration`) is captured. This is largely benign for ordinary `.ts`, but it is a real gap for `.d.ts` declaration files, which consist almost entirely of bodyless `declare` and signature forms; an ambient API surface defined that way is invisible to `sem`.

## Reproduction

```sh
cat > types.d.ts <<'EOF'
export declare function createClient(opts: ClientOptions): Client;
export declare class Client {
  connect(): Promise<void>;
}
export interface ClientOptions { host: string; }
export declare const VERSION: string;
export type Row = Record<string, unknown>;
EOF
cat > overloads.ts <<'EOF'
export function combine(a: string, b: string): string;
export function combine(a: number, b: number): number;
export function combine(a: any, b: any): any { return a + b; }
declare function ambientFn(x: number): number;
EOF

sem entities types.d.ts
sem entities overloads.ts
```

## Observed

`sem entities types.d.ts` lists `class Client` (with `connect`), `interface ClientOptions`, `variable VERSION`, and `type Row` — but `createClient` (the `declare function`) is missing. `declare class`, `interface`, `const`, and `type` all extract correctly; only the standalone ambient function is lost.

`sem entities overloads.ts` lists a single `function combine` (the implementation at L3 only). Both overload heads (L1, L2) and `declare function ambientFn` (L4) are dropped.

## Expected

Ambient and declared function signatures are extractable, so a `.d.ts` API surface and overload contracts are visible in `sem entities`, `sem diff`, and `sem context`.

## Root cause

`crates/sem-core/src/parser/plugins/code/languages.rs`: `function_signature` is absent from `TYPESCRIPT_CONFIG`/`TSX_CONFIG` `entity_node_types` (L412–426 and L439–453). The type mapping itself is already in place — `crates/sem-core/src/parser/plugins/code/entity_extractor.rs` `map_node_type` maps `function_signature` to `"function"` at L1551 — so only the `entity_node_types` configuration is missing. The grammar (`tree-sitter-typescript` 0.23.2) emits `function_signature` for both overload heads and `declare function`.

## Suggested fix

This carries a tradeoff worth handling deliberately rather than naively. Adding `function_signature` to `entity_node_types` unconditionally would also emit every overload head as a separate `function` entity alongside the implementation. Those duplicates would be line-disambiguated by the existing post-pass in `entity_extractor.rs:37-58` (which appends `@L{line}` via `build_entity_id_disambiguated`), so they would not collide, but they would be noisy — a single overloaded `combine` would surface as three `combine` entities.

A targeted approach: extract `function_signature` only when it has no sibling `function_declaration` implementation of the same name in scope (i.e. genuine ambient declarations and `.d.ts` content), or fold overload heads into the implementation entity's signature rather than emitting them separately. Add tests for a `.d.ts` `declare function` and for an overloaded function with multiple heads plus one implementation.

## Related issues

Related to #151 (TypeScript interface method signatures); the bodyless `function_signature` form (overload heads and `declare function`) was not covered by that work.

Iron-Ham:
Fixes #265

## Summary
- extract TypeScript/TSX `function_signature` nodes so ambient `declare function` declarations appear as function entities
- suppress overload signature entities when a same-scope implementation exists
- cache implementation names per TypeScript scope to avoid repeated sibling scans in large declaration files

## Test plan
- `cargo test -p sem-core`
- `cargo test -p sem-cli`
- dummy git repos under `/tmp/sem-issue-265-*` validating `sem entities` and `sem diff` for ambient declarations, overload de-duplication, nested namespace overloads, mixed export overloads, and declaration-file signature changes


Iron-Ham:
Superseded by #309.

PR #293, opened 2026-06-01
Make verify cache bypassable and consistent

### LINKED ISSUE #250: sem verify always uses the cache and has no --no-cache, producing CI false negatives
**Severity:** High
**Component:** cache / verify command
**Affected commands:** `sem verify`

## Summary

`sem verify` is hardwired to read the incremental cache and exposes no `--no-cache` flag. When the cache is stale (see the incremental-cache divergence filed alongside this), a genuine arity bug is missed: `verify` prints an empty result and exits 0 when it should report the mismatch and exit 1. For a command whose job is to gate CI, a silent false negative is the worst failure mode, and there is no escape hatch to force a correct run.

## Reproduction

```sh
cat > b.py <<'EOF'
def other():
    return 0
EOF
cat > a.py <<'EOF'
def use_it():
    return target(1, 2, 3)
EOF

sem graph --json >/dev/null    # prime the cache while target() does not exist
sleep 1.1

# add target(a) — a takes ONE param, but use_it calls it with THREE
cat > b.py <<'EOF'
def other():
    return 0

def target(a):
    return a
EOF

sem verify --json ; echo "exit=$?"   # stale cache
rm -rf .sem
sem verify --json ; echo "exit=$?"   # fresh
sem verify --no-cache                # rejected
```

## Observed

With the stale cache, `sem verify --json` prints `[]` and exits 0. After `rm -rf .sem`, `sem verify --json` correctly reports the arity mismatch (`use_it` calls `target` with 3 args; `target` expects 1) and exits 1. `sem verify --no-cache` fails with `error: unexpected argument '--no-cache' found`.

## Expected

`sem verify` should report the mismatch and exit 1 in both cases. At minimum it should accept `--no-cache` like every other graph-reading command, and ideally default to a fresh build given its gating role.

## Root cause

`verify_command` in `crates/sem-cli/src/commands/verify.rs:26` calls `get_or_build_graph(root, &file_paths, &registry, false)` — the trailing `false` is the `no_cache` argument, hardcoded, so `verify` always uses the cache. `VerifyOptions` (`crates/sem-cli/src/commands/verify.rs:6-12`) has no `no_cache` field and the CLI parser rejects the flag. The underlying false negative is the incremental-cache divergence: a newly-added cross-file call edge is never re-resolved into the cache, so `find_arity_mismatches` never sees the call site.

## Suggested fix

Add a `--no-cache` flag to `verify` and thread it into `get_or_build_graph`; consider defaulting `verify` to a fresh build. Independently, fix the incremental-cache divergence so the cached path is trustworthy.

## Related issues

Depends on a separate issue covering the incremental-cache divergence that drops cross-file call edges. Related to #123 (closed, "sem verify produces too many false positives") and #209 (closed, verify first call site).

Iron-Ham:
## Summary
- Add `--no-cache` support to `sem verify`.
- Re-resolve clean callers that mention changed stale-file callees so incremental graph rebuilds can add previously missing call edges.
- Bump the shared SQLite cache schema version to invalidate pre-fix graph snapshots.
- Persist complete entity snapshots during incremental cache saves so rewritten edges cannot point at repaired clean entity IDs missing from the cache.

Fixes #250

## Test plan
- `cargo test --manifest-path crates/Cargo.toml -p sem-cli --test verify_cli -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-cli cache::tests::open_rebuilds_cache_when_schema_version_is_unsupported -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-mcp cache::tests::open_rebuilds_cache_when_schema_version_is_unsupported -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml --workspace -- --nocapture`
- Manual dummy git repos covering incremental stale-cache verification and seeded v1 bad full-cache invalidation.

### DIFF
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -386,13 +386,8 @@ impl DiskCache {
             .iter()
             .filter(|file| !shared_cache::is_manifest_file_name(file))
             .collect();
-        let stale_set: HashSet<&str> = source_stale_files
-            .iter()
-            .map(|file| file.as_str())
-            .collect();
 
         let tx = self.conn.unchecked_transaction()?;
-        let mut deleted_cached_files = Vec::new();
 
         // Delete stale file entries
         {
@@ -420,7 +415,6 @@ impl DiskCache {
                     && !current_set.contains(path.as_str())
                 {
                     del_files.execute(params![path])?;
-                    deleted_cached_files.push(path.clone());
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


Iron-Ham:

PR #295, opened 2026-06-01
Fix same-line entity ID collisions

### LINKED ISSUE #246: Entity IDs collide when two sibling entities share a start line, silently dropping changes
**Severity:** Critical
**Component:** parser / entity identity (entity ID disambiguation, reorder detection)
**Affected commands:** `sem diff`, `sem graph`, `sem impact`, `sem context`, `sem log` (anything keyed by entity ID)

## Summary

The collision-disambiguation suffix appended to entity IDs is `@L{start_line}`, which is not collision-proof: two entities with the same name, type, and parent that also share a `start_line` (overloads on one line, minified or transpiler-generated single-line files) receive an identical suffix and re-collide to one ID. Because the differ and the graph key entities by ID, one of the colliding entities is silently dropped — so a deletion or modification of that entity can be reported as "no change." This is silent incorrectness in `sem`'s core promise. A second manifestation of the same shared-`start_line` weakness produces spurious "reordered" results.

### Finding A — same-line overloads collapse; a deletion becomes invisible

```sh
printf 'function f(a: number): void {}; function f(a: string): void {}\n' > over.ts

# entities sees two `f`; graph keyed by id sees only one
sem entities over.ts
sem graph over.ts --json

# delete the second overload, then diff
printf 'function f(a: number): void {}\n' > over.ts
sem diff over.ts ; echo "exit=$?"
```

### Finding B — spurious "reordered" for unchanged entities sharing a start line

```sh
printf 'function a(){return 1} function b(){return 2} function c(){return 3} function d(){return 4}\n' > min.js
# edit only c(); a, b, d are untouched and in the same order
printf 'function a(){return 1} function b(){return 2} function c(){return 999} function d(){return 4}\n' > min.js
sem diff min.js --json
```

### Finding A

`sem entities over.ts` lists two `function f`, both at `L1:1`. `sem graph over.ts --json` reports `entityCount = 1` with a single id `over.ts::function::f@L1` — one overload was dropped. After deleting the second overload, `sem diff over.ts` prints:

```
No semantic changes detected.
exit=0
```

The deletion is completely invisible (`--json` summary shows `deleted=0, total=0`).

### Finding B

`sem diff min.js --json` reports `[modified c, reordered a, reordered b]` (`reordered` count = 2) even though only `c` changed and nothing moved. The same edit with one function per line correctly reports only `[modified c]`.

## Expected

Two same-line overloads must be tracked as two distinct entities; deleting one must be reported as a deletion. Unchanged entities that merely share a `start_line` must not be reported as reordered.

## Root cause

The disambiguation post-pass in `extract_entities` (`crates/sem-core/src/parser/plugins/code/entity_extractor.rs:37-58`) re-assigns colliding IDs using `build_entity_id_disambiguated(file, type, name, parent, start_line)` — the only added discriminator is `start_line`. Two entities identical in name/type/parent and on the same line get the same suffix and re-collide.

Every ID-keyed consumer then drops one of the colliding entities:

- `crates/sem-core/src/model/identity.rs:161-164` builds `before_by_id` / `after_by_id` as `HashMap<&str, &SemanticEntity>` keyed by `id.as_str()`; duplicate IDs collapse (last write wins), so the differ never sees the second entity.
- `crates/sem-core/src/parser/graph.rs:182-193` inserts into `entity_map` keyed by id, collapsing the same way.

Finding B is a second symptom of the shared-`start_line` weakness. `detect_reorders` (`crates/sem-core/src/model/identity.rs:482-503`) runs a longest-increasing-subsequence over the after-side `start_line` values, and the LIS uses a strict-`<` comparator: `tails.partition_point(|&t| t < seq[i])` (`crates/sem-core/src/model/identity.rs:522`). When many entities share `start_line` (e.g. `[1,1,1,1]`), no equal element can extend the subsequence, so the LIS collapses to length 1 and every other equal-line entity is flagged `reordered`.

## Suggested fix

- Make the disambiguator collision-proof: when `start_line` ties within a collision group, append a per-group ordinal (`@L{line}#{n}`) or a content-hash suffix, and assert ID uniqueness at the end of the post-pass in `entity_extractor.rs`.
- In reorder detection, use a non-strict comparator with a stable secondary tie-break (column or byte offset), or skip reorder detection for entities that share a `start_line`.

### Downstream note

Hotspot churn keys entities on `(name, type, file)` in `crates/sem-core/src/parser/hotspot.rs:41` (declaration) and `:70-74` (construction), with no line discriminator at all, so same-line entities are conflated in churn counts as well. Fixing the identity collision upstream limits the blast radius, but the hotspot key should also include a line/ordinal discriminator.

## Related issues

Incomplete fix of #140 (closed, "Same-name overloads collapse/disambiguate", whose `@L{start_line}` disambiguation was added by #164 / #165) — entities that share a `start_line` re-collide despite that work. Related to #115 (closed, nested entity IDs duplicate the file path).

Iron-Ham:
Fixes #246

## Summary
- Disambiguate colliding code entity IDs with line and ordinal suffixes, including same-line siblings.
- Propagate rewritten parent IDs to descendants so graph parent links stay valid.
- Preserve identity matching across collision-group shrink/growth and avoid false reorder reports for unchanged same-line entities.
- Include entity IDs in hotspot rows and key hotspot churn by entity identity.
- Add parser, identity, and CLI regressions for overload deletion, survivor edits, insertion, reorder detection, and duplicate parent propagation.

## Test plan
- `cargo test -p sem-core same_line -- --nocapture`
- `cargo test -p sem-core collision_group -- --nocapture`
- `cargo test -p sem-core duplicate_parent -- --nocapture`
- `cargo test -p sem-cli --test diff_file_compare same_line -- --nocapture`
- `cargo test --workspace`
- Validated against disposable git repos for overload deletion, shrink+edit, grow+edit, minified edit without reorder noise, same-line swap reorder, and duplicate same-line parent propagation.

### DIFF
--- a/crates/sem-cli/tests/diff_file_compare.rs
+++ b/crates/sem-cli/tests/diff_file_compare.rs
@@ -22,6 +22,32 @@ fn run_sem_json(dir: &PathBuf, home: &PathBuf, args: &[&str]) -> std::process::O
         .expect("sem should run")
 }
 
+fn run_git(dir: &PathBuf, args: &[&str]) {
+    let output = Command::new("git")
+        .args(args)
+        .current_dir(dir)
+        .output()
+        .expect("git should run");
+    assert!(
+        output.status.success(),
+        "git {:?} failed\nstdout: {}\nstderr: {}",
+        args,
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr)
+    );
+}
+
+fn init_repo(dir: &PathBuf) {
+    run_git(dir, &["init", "-q"]);
+    run_git(dir, &["config", "user.email", "a@b.co"]);
+    run_git(dir, &["config", "user.name", "a"]);
+}
+
+fn commit_all(dir: &PathBuf, message: &str) {
+    run_git(dir, &["add", "-A"]);
+    run_git(dir, &["commit", "-qm", message]);
+}
+
 #[test]
 fn cross_language_file_compare_uses_each_side_path() {
     let dir = temp_dir("cross-language-file-compare");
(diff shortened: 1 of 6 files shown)


Iron-Ham:
Superseded by #312.

PR #300, opened 2026-06-01
Harden patch diff parsing

### LINKED ISSUE #271: `sem diff --patch` parsing gaps: quoted non-ASCII headers, `" b/"` paths, and ignored hunk bodies
**Severity:** Medium
**Component:** sem diff --patch (unified-diff parsing in `parse_unified_diff`)
**Affected commands:** `sem diff --patch`

## Summary

The unified-diff parser behind `sem diff --patch` has three robustness gaps. It fails outright on git's default quoted output for non-ASCII filenames, which breaks the documented `git diff | sem diff --patch` workflow out of the box. It mis-extracts the file path when the path itself contains the substring `" b/"`. And it reconstructs file contents only from blob SHAs via `git show`, ignoring the `+`/`-` hunk lines entirely, so patches without an `index` line (or whose blobs are not in the local object database) silently yield zero or wrong changes.

### Finding A — fails on git's default quoted non-ASCII headers

```sh
printf 'def foo():\n    return 1\n' > "ünïcödé.py"
git add "ünïcödé.py" && git commit -qm c1
printf 'def foo():\n    return 2\n' > "ünïcödé.py" && git add "ünïcödé.py"

git diff --cached | sem diff --patch --json                    # default git config
git -c core.quotepath=false diff --cached | sem diff --patch --json  # workaround
```

### Finding B — path containing `" b/"` is truncated

```sh
mkdir -p "my b"
printf 'def foo():\n    return 1\n' > "my b/app.py"
git add "my b/app.py" && git commit -qm c1
printf 'def foo():\n    return 2\n' > "my b/app.py" && git add "my b/app.py"

git -c core.quotepath=false diff --cached | sem diff --patch --json
# inspect changes[0].filePath
```

### Finding C — hunk bodies are ignored; index-less patches yield zero changes

```sh
cat > plain.patch <<'PATCH'
diff --git a/app.py b/app.py
--- a/app.py
+++ b/app.py
@@ -1,2 +1,2 @@
 def foo():
-    return 1
+    return 2
PATCH

sem diff --patch < plain.patch --json   # a clear modification, but no `index` line
```

### Finding A

With git's default configuration (`core.quotepath` unset, which defaults to true), the header is emitted as `diff --git "a/\303\274n..." "b/..."` and sem fails:

```text
error: no recognizable diff hunks in stdin (expected 'diff --git' headers and '@@ ... @@' hunk markers)   (exit 1)
```

`core.quotepath=false` is required to make it work.

### Finding B

`changes[0].filePath` is `app.py` instead of `my b/app.py`.

### Finding C

`sem diff --patch < plain.patch` reports zero changes plus a warning, and exits 0:

```text
warning: could not resolve contents for app.py. Try running from inside the repo, or use -C /path/to/repo.
0
```

A related variant — a patch that *does* carry an `index` line whose blobs are absent from the local object DB (email/review/cross-repo patches) — does not yield zero but a wrong result: the working-tree fallback reads the on-disk file and reports its current content as `added` while ignoring the patch's `+` side entirely.

## Expected

`git diff | sem diff --patch` should work without requiring `core.quotepath=false`, since that workflow is documented and quoting is git's default for non-ASCII names (Finding A). The path should be parsed as `my b/app.py` (Finding B). A patch that plainly modifies `foo()` should report one modification, reconstructing content from the hunk lines when blob resolution is unavailable, or fail loudly rather than emitting empty/misleading success (Finding C).

## Root cause

All three live in `parse_unified_diff` in `crates/sem-cli/src/commands/diff.rs`:

- **Finding A:** the header is matched with `line.strip_prefix("diff --git a/")` (`crates/sem-cli/src/commands/diff.rs:474`), which only recognizes the unquoted form; quoted headers (`diff --git "a/...`) never match, so the entry is skipped and the parser reports no recognizable hunks.
- **Finding B:** the b-side path is extracted with `rest.rfind(" b/")` (`crates/sem-cli/src/commands/diff.rs:484`), a last-occurrence search; for `a/my b/app.py b/my b/app.py` the trailing `" b/"` wins and the path is truncated to `app.py`.
- **Finding C:** content is reconstructed solely from the `index` blob SHAs via `git show`, with a working-tree fallback for the new side (`crates/sem-cli/src/commands/diff.rs:586-630`); the `+`/`-` hunk lines are never read. With no resolvable `index` line, both `before_content` and `after_content` are `None`, producing an empty change set with only a warning.

## Suggested fix

- Detect a leading `"` on the `diff --git` header and C-style-unquote both sides before parsing the paths (Finding A).
- Parse the `--- a/` and `+++ b/` header lines (handling `/dev/null` and quoting) to obtain the paths, rather than searching the `diff --git` line (Findings A and B).
- When blob resolution is unavailable, reconstruct before/after content from the hunk bodies, or fail loudly with a clear error instead of emitting empty success (Finding C).

## Related issues

Builds on #223 (merged, "Validate patch input for sem diff", closing #187); that validation rejects malformed input but does not address quoted non-ASCII headers, the `" b/"` path case, or hunk-body reconstruction.

Iron-Ham:
## Summary

- decode Git-quoted path tokens in patch headers and binary metadata
- preserve paths containing ` b/` by using file headers and safer diff-header splitting
- reconstruct before/after content from hunk bodies when patch blobs are unavailable
- keep local worktree fallback only when the file hash matches the patch target object

Closes #271

## Test plan

- `rustfmt --check sem-cli/src/commands/diff.rs sem-cli/tests/diff_patch.rs`
- `cargo test -p sem-cli parse_unified_diff`
- `cargo test -p sem-cli`
- Manual dummy repositories covering quoted non-ASCII paths, ` b/` paths, indexless patches, and partially resolvable index patches

### DIFF
--- a/crates/sem-cli/src/commands/diff.rs
+++ b/crates/sem-cli/src/commands/diff.rs
@@ -232,7 +232,6 @@ fn normalize_trailing_output_format(opts: &mut DiffOptions) {
     opts.args = args;
 }
 
-
 fn parse_args(args: Vec<String>, cwd: &str) -> ParsedArgs {
     let (refs, pathspecs) = split_on_separator(args);
 
@@ -346,7 +345,7 @@ fn parse_args(args: Vec<String>, cwd: &str) -> ParsedArgs {
 }
 
 /// Parse a unified diff (e.g. from `git diff`) into FileChange entries.
-/// Uses blob SHAs from `index` lines to retrieve full file contents via `git show`.
+/// Uses blob SHAs when possible and hunk bodies when the patch is not tied to local blobs.
 #[derive(Debug, PartialEq, Eq)]
 enum PatchParseError {
     EmptyInput,
@@ -408,7 +407,26 @@ fn is_binary_files_marker(line: &str) -> bool {
     let Some(paths) = rest.strip_suffix(" differ") else {
         return false;
     };
-    let Some((old, new)) = paths.split_once(" and ") else {
+
+    let Some((old, new)) = (if paths.trim_start().starts_with('"') {
+        let Some((old, rest)) = parse_git_path_token(paths) else {
+            return false;
+        };
+        let Some(rest) = rest.strip_prefix(" and ") else {
+            return false;
+        };
+        let Some((new, rest)) = parse_git_path_token(rest) else {
+            return false;
+        };
+        if !rest.is_empty() {
+            return false;
+        }
+        Some((old, new))
+    } else {
+        paths
+            .split_once(" and ")
+            .map(|(old, new)| (old.to_string(), new.to_string()))
+    }) else {
         return false;
     };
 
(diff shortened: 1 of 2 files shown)


Iron-Ham:

PR #301, opened 2026-06-01
Preserve incremental graph edges for stale targets

### LINKED ISSUE #249: Incremental cache drops a clean caller's edge into a content-clean entity in an edited file
**Severity:** High
**Affected commands:** `sem impact`, `sem graph`, `sem context`, `sem verify` (everything that reads the cached graph)

## Summary

After an everyday edit — adding a sibling function to a file — the incremental cache produces different results than a fresh `--no-cache` build. An edge from a clean caller to a target that is byte-identical but whose file became stale (because some other entity in that file changed) is dropped and never regenerated. The stale result persists across runs, so `sem impact`, `sem graph`, and `sem verify` silently return incomplete dependency data.

## Reproduction

```sh
printf 'def helper():\n    return 1\n' > b.py
printf 'def use_it():\n    return helper()\n' > a.py

sem graph --json >/dev/null      # prime the cache: edge use_it -> helper
sleep 1.1                          # ensure a distinct mtime

# add an unrelated function to b.py; helper() stays byte-identical
printf 'def helper():\n    return 1\n\ndef unrelated():\n    return 42\n' > b.py

sem impact helper --json            # cached
sem impact helper --json --no-cache # fresh build
```

## Observed

Cached `sem impact helper` reports no dependents (`[]`); the fresh `--no-cache` build reports `['use_it']`. The two diverge. Editing `helper` itself, or making a cosmetic-only edit inside `helper`, does not diverge — only adding/removing a sibling entity in the same file triggers it.

## Expected

The cached result must match the fresh build: `use_it` is a dependent of `helper` regardless of unrelated edits in `helper`'s file.

## Root cause

In `EntityGraph::build_incremental` (`crates/sem-core/src/parser/graph.rs`):

- `kept_edges` (`crates/sem-core/src/parser/graph.rs:603-623`) drops every cached edge whose `to_entity` lives in a stale file (`to_stale` is true at `:614`), even when that target entity is content-clean. The edge `use_it -> helper` is dropped because `helper`'s file `b.py` is stale.
- `affected_clean_ids` (`crates/sem-core/src/parser/graph.rs:573-586`) only marks a clean caller for re-resolution when its cached edge pointed at a *truly-changed* or *deleted* target (`:576-577`). Since `helper` is content-clean (only its file changed), `use_it` is never added, so the dropped edge is never regenerated.

The caller's file mtime is unchanged, so on subsequent runs the caller hits the full-cache path and the missing edge persists indefinitely.

## Suggested fix

In `affected_clean_ids`, also mark for re-resolution any clean caller whose cached edge points into a stale file (not just edges into truly-changed/deleted targets); or, in `kept_edges`, retain edges whose `to_entity` is a content-clean entity that happens to live in a stale file.

## Related issues

Side-effect of the fix for #116 (closed, "Incremental graph cache keeps stale edges when import scope changes") — the `kept_edges` over-drop is deliberate (the comment at `crates/sem-core/src/parser/graph.rs:600-602` cites that issue), but it drops a clean caller's edge into a content-clean entity in a stale file and never regenerates it. Related to #144 and #205.

Iron-Ham:
## Summary
- Re-resolve clean callers whose cached edges point into stale files so content-clean stale targets keep their dependents.
- Re-resolve clean files that import stale files when added stale-file entities can start resolving clean references, including aliases and namespace imports.
- Bump the shared cache schema version so persisted graph caches rebuild.

Fixes #249

## Test plan
- cargo test -p sem-core
- cargo test -p sem-cli
- cargo test -p sem-mcp
- cargo build -p sem-cli
- Disposable git repos comparing cached `sem impact` output with `--no-cache` for Python and TypeScript content-clean, added-target, alias, and namespace scenarios.

### DIFF
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -737,7 +737,11 @@ mod tests {
 
     #[test]
     fn open_rebuilds_cache_when_schema_version_is_unsupported() {
-        for version in [0, shared_cache::CACHE_SCHEMA_VERSION + 1] {
+        for version in [
+            0,
+            shared_cache::CACHE_SCHEMA_VERSION - 1,
+            shared_cache::CACHE_SCHEMA_VERSION + 1,
+        ] {
             let root = temp_repo_root(&format!("unsupported-{version}"));
             seed_unsupported_cache(&root, version);
 
--- a/crates/sem-core/src/parser/graph.rs
+++ b/crates/sem-core/src/parser/graph.rs
@@ -28,7 +28,7 @@ macro_rules! maybe_par_iter {
 
 use crate::git::types::{FileChange, FileStatus};
 use crate::model::entity::SemanticEntity;
-use crate::parser::import_resolution::find_import_target;
+use crate::parser::import_resolution::{find_import_target, import_source_matches_file};
 use crate::parser::registry::{resolve_go_method_parent_ids, ParserRegistry};
 use crate::parser::scope_resolve;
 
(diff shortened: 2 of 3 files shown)


Iron-Ham:

PR #302, opened 2026-06-01
Extract Swift multi-binding properties

### LINKED ISSUE #260: Swift multi-binding property declaration drops all but the first name
**Severity:** Low
**Affected commands:** `sem entities`, `sem impact`, `sem context`, `sem blame`

## Summary

A Swift multi-binding property declaration such as `var x, y: Int` extracts only the first name (`x`); the remaining names are not emitted as entities. The trailing bindings are therefore not independently addressable by `sem impact`/`sem context`/`sem blame`, though a change to the shared declaration still registers against the first entity.

## Reproduction

```sh
cat > Example.swift <<'EOF'
struct Point {
    var x, y: Int
}
EOF
sem entities Example.swift --json | jq -r '.[] | [.type, .name] | @tsv'
```

## Observed

Only `x` is extracted; `y` is absent.

```
struct    Point
property  x
```

## Expected

Both `x` and `y` should be represented (one entity each), or the declaration should be handled in a way that does not silently drop `y`.

## Root cause

`extract_name` (`crates/sem-core/src/parser/plugins/code/entity_extractor.rs:1174`) returns a single name per node, and entity emission produces one entity per matching node (`entity_extractor.rs:431-446`). There is no per-binding expansion for a Swift `property_declaration` that declares multiple names, so only the first binding becomes an entity.

## Suggested fix

In the Swift path, expand a multi-binding `property_declaration` into one entity per declared name.

## Related issues

Same class as #149 ("Multi-declarator declarations collapse to the first entity", JS/TS); the Swift `var x, y: Int` form was not covered there.

Iron-Ham:
## Summary
- emit one Swift property entity per binding for declarations like `var x, y: Int`
- synthesize per-binding content for shared type annotations, explicit per-binding types, initializers, and `let` declarations
- use extractor sequence positions for reorder detection so same-line entities do not produce false reorders

Fixes #260

## Test plan
- `cargo test -p sem-core swift_multi_binding --features lang-swift -- --nocapture`
- `cargo test -p sem-core reorder --features lang-swift -- --nocapture`
- `cargo test -p sem-core --features lang-swift`
- `cargo test -p sem-cli`
- `cargo build -p sem-cli`
- disposable Swift git repo validating `sem entities`, `sem context`, `sem impact`, `sem blame`, and `sem diff --format json`

### DIFF
--- a/crates/sem-core/src/model/identity.rs
+++ b/crates/sem-core/src/model/identity.rs
@@ -452,10 +452,21 @@ fn detect_reorders(
     // Collect unchanged entities: matched by ID with same content_hash
     let before_by_id: HashMap<&str, &SemanticEntity> =
         before.iter().map(|e| (e.id.as_str(), e)).collect();
+    let before_positions: HashMap<&str, usize> = before
+        .iter()
+        .enumerate()
+        .map(|(i, e)| (e.id.as_str(), i))
+        .collect();
+    let after_positions: HashMap<&str, usize> = after
+        .iter()
+        .enumerate()
+        .map(|(i, e)| (e.id.as_str(), i))
+        .collect();
 
     // Group by file. For each file, collect unchanged entities in their
     // before-order, then look up their after-positions.
-    let mut by_file: HashMap<&str, Vec<(&SemanticEntity, &SemanticEntity)>> = HashMap::new();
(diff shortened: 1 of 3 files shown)


Iron-Ham:
Superseded by #310.

PR #305, opened 2026-06-01
Extract TypeScript abstract entities

### LINKED ISSUE #262: TypeScript `abstract class` and its abstract methods are never extracted as entities
**Severity:** High
**Affected commands:** `sem entities`, `sem diff`, `sem impact`, `sem context`, `sem blame`

## Summary

A TypeScript `abstract class` parses as the tree-sitter node `abstract_class_declaration`, which is absent from the TypeScript/TSX `entity_node_types`, so the class is never emitted as an entity. Its concrete methods are still reached by the generic child walk and emitted as top-level orphans (`parent_id: null`), and its abstract method signatures (`abstract_method_signature`) are dropped entirely. Because there is no graph vertex for the class, `sem impact` and `sem context` on it fail outright, and callers of its methods are not linked. Abstract classes are pervasive in real TypeScript (NestJS providers, Angular base components, TypeORM base entities), so this is a significant gap against the documented "classes" support.

This report covers two issues that share one root cause and one fix: the missing class node type (Finding A) and the missing abstract-method node type (Finding B).

### Finding A — `abstract_class_declaration` not extracted

The class entity is missing; its concrete methods are orphaned.

### Finding B — `abstract_method_signature` not extracted

Even after Finding A is fixed, abstract method declarations (`abstract area(): number;`) remain invisible, because their node kind is also not configured. Abstract methods are the contract the class defines, so they are exactly what subclasses and callers care about.

## Reproduction

```sh
cat > shapes.ts <<'EOF'
export abstract class Shape {
  abstract area(): number;
  describe(): string {
    return `area is ${this.area()}`;
  }
}

export class Circle extends Shape {
  constructor(private radius: number) { super(); }
  area(): number { return Math.PI * this.radius * this.radius; }
}
EOF
cat > main.ts <<'EOF'
import { Shape, Circle } from "./shapes";
export function run(): void {
  const shapes: Shape[] = [new Circle(2)];
  for (const s of shapes) { console.log(s.describe()); }
}
EOF

sem entities shapes.ts
sem impact Shape --file shapes.ts --no-cache
sem context Shape --file shapes.ts
sem impact describe --file shapes.ts --no-cache
```

## Observed

`sem entities shapes.ts`:

```
  method describe (L3:5)
  class Circle (L8:11)
    method constructor (L9:9)
    method area (L10:10)
```

The `--json` form confirms `describe` has `"parent_id": null` while `Circle`'s own methods correctly carry `"parent_id": "shapes.ts::class::Circle"`. `class Shape` and the abstract `area()` are entirely absent.

`sem impact Shape` and `sem context Shape` both print `error: Entity 'Shape' not found`. `sem impact describe` reports `No other entities are affected`, even though `main.ts` calls `s.describe()`.

A control with the same hierarchy minus the `abstract` keyword behaves correctly: the concrete base class is extracted, `impact` lists the subclass and the cross-file caller, and method-call-through-declared-type resolution links the caller to the base method. The only differing input is the `abstract` keyword, which isolates the cause.

## Expected

`abstract class Shape` is emitted as a `class` entity with `area` and `describe` nested under it. `sem impact Shape` lists `Circle` and the cross-file caller; `sem context Shape` returns context; the abstract `area()` appears as a (abstract) method.

## Root cause

`crates/sem-core/src/parser/plugins/code/languages.rs`: `TYPESCRIPT_CONFIG.entity_node_types` (L412–426) and `TSX_CONFIG.entity_node_types` (L439–453) list `class_declaration` but neither `abstract_class_declaration` nor `abstract_method_signature`. In `crates/sem-core/src/parser/plugins/code/entity_extractor.rs`, `map_node_type` maps `class_declaration` to `"class"` at L1555; `abstract_class_declaration` and `abstract_method_signature` are unmapped and would fall through to the `other => other` arm at L1593.

There is an internal inconsistency: the same `languages.rs` file's `TS_SCOPE_CONFIG` already lists `abstract_class_declaration` in `class_scope_nodes` (L1000) and in `init_strategy.class_nodes` (L1030). The scope-resolution layer handles abstract classes; the entity-extraction layer does not match.

The grammar (`tree-sitter-typescript` 0.23.2) confirms `abstract_class_declaration` has a `name` field (`type_identifier`) and a `body` field (`class_body`, already a container type), so name extraction and method nesting will work as soon as the node is treated as an entity. `abstract_method_signature` likewise has a `name` field.

## Suggested fix

1. Add `"abstract_class_declaration"` and `"abstract_method_signature"` to `entity_node_types` in both `TYPESCRIPT_CONFIG` and `TSX_CONFIG`.
2. In `map_node_type`, add `"abstract_class_declaration"` to the `… => "class"` arm (L1555) and `"abstract_method_signature"` to the `… => "method"` arm (L1553).
3. Add a regression test asserting that `abstract class Foo { abstract m(): void; n(): void {} }` yields a `class Foo` entity with `m` and `n` nested under it.

## Related issues

Same "missing node type" class as #150/#151/#152; abstract classes and abstract method signatures were not covered by that work.

Iron-Ham:
## Summary
- Extract TypeScript and TSX abstract classes and abstract method signatures as class/method entities
- Resolve typed TypeScript for-of loop variables for Shape[] and Array<Shape>-style annotations so impact/context include method callers
- Keep loop variable bindings byte-scoped to loop bodies and avoid leaking nested block declarations into function scope

Fixes #262

## Test Plan
- cargo test --manifest-path crates/Cargo.toml -p sem-core
- cargo test --manifest-path crates/Cargo.toml -p sem-cli
- cargo build --manifest-path crates/Cargo.toml -p sem-cli
- git diff --no-ext-diff --check
- Disposable git repos validating sem entities, sem impact, and sem context for .ts, .tsx, Shape[], Array<Shape>, nested for-of, and leakage regressions

## Review
- Independent subagent review found a block-scoped declaration leak; fixed by separating nested for-of scanning from normal assignment scanning and adding regression coverage
- Follow-up independent subagent review found no blocking issues


Iron-Ham:
Superseded by #309.

PR #306, opened 2026-06-01
Consolidate diff path and patch parsing

### LINKED ISSUE #269: `sem diff --patch` pathspec matching diverges from git (directory over-match, no bracket classes)

### LINKED ISSUE #270: `sem diff <ref> <path>` is parsed as a revision range, unlike `git diff`

### LINKED ISSUE #271: `sem diff --patch` parsing gaps: quoted non-ASCII headers, `" b/"` paths, and ignored hunk bodies

### LINKED ISSUE #274: `sem diff --json` emits two different JSON shapes for an empty result

Iron-Ham:
Summary
- Consolidates empty diff JSON output, patch pathspec matching, diff ref/path argument parsing, and hardened patch parsing.
- Keeps the changes scoped to diff path/patch behavior and parser output shape.

Supersedes #280, #281, #284, #300.
Closes https://github.com/Ataraxy-Labs/sem/issues/274
Closes https://github.com/Ataraxy-Labs/sem/issues/269
Closes https://github.com/Ataraxy-Labs/sem/issues/270
Closes https://github.com/Ataraxy-Labs/sem/issues/271

Test plan
- cargo test -p sem-cli --test diff_patch
- cargo test -p sem-cli commands::diff::tests
- cargo test -p sem-cli --test diff_file_compare

### DIFF
--- a/crates/sem-cli/src/commands/diff.rs
+++ b/crates/sem-cli/src/commands/diff.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::HashSet;
 use std::io::Read;
 use std::path::Path;
@@ -78,12 +79,112 @@ fn split_on_separator(args: Vec<String>) -> (Vec<String>, Vec<String>) {
     }
 }
 
-/// Simple glob matching for pathspecs. Supports `*` (any chars except `/`),
-/// `**` (any chars including `/`), and `?` (one non-`/` char).
+fn looks_like_pathspec(arg: &str) -> bool {
+    arg.starts_with(':') || arg.contains('*') || arg.contains('?') || arg.contains('[')
+}
+
+fn pathspec_like_arg_is_valid_rev(
+    arg: &str,
+    git: &mut Option<Option<GitBridge>>,
+    cwd: &str,
+) -> bool {
+    arg.starts_with(':') && is_valid_git_rev(git, cwd, arg)
+}
+
+fn get_git_bridge<'a>(git: &'a mut Option<Option<GitBridge>>, cwd: &str) -> Option<&'a GitBridge> {
+    if git.is_none() {
+        *git = Some(GitBridge::open(Path::new(cwd)).ok());
+    }
+
+    git.as_ref().and_then(|git| git.as_ref())
+}
+
+fn is_valid_git_rev(git: &mut Option<Option<GitBridge>>, cwd: &str, arg: &str) -> bool {
+    get_git_bridge(git, cwd).is_some_and(|git| git.is_valid_rev(arg))
+}
+
+/// Simple glob matching for pathspecs. Supports `*`, `**`, `?`, and bracket classes.
 fn glob_match(pattern: &str, path: &str) -> bool {
+    if pattern.is_ascii() && path.is_ascii() {
+        return glob_match_inner_bytes(pattern.as_bytes(), path.as_bytes());
+    }
+
     let pat: Vec<char> = pattern.chars().collect();
     let text: Vec<char> = path.chars().collect();
     glob_match_inner(&pat, &text)
+}
+
+fn posix_bracket_class_matches(name: &str, ch: char) -> Option<bool> {
+    match name {
+        "alnum" => Some(ch.is_ascii_alphanumeric()),
+        "alpha" => Some(ch.is_ascii_alphabetic()),
+        "blank" => Some(ch == ' ' || ch == '\t'),
+        "cntrl" => Some(ch.is_ascii_control()),
+        "digit" => Some(ch.is_ascii_digit()),
+        "graph" => Some(ch.is_ascii_graphic()),
+        "lower" => Some(ch.is_ascii_lowercase()),
+        "print" => Some(ch.is_ascii_graphic() || ch == ' '),
+        "punct" => Some(ch.is_ascii_punctuation()),
+        "space" => Some(ch.is_ascii_whitespace()),
+        "upper" => Some(ch.is_ascii_uppercase()),
+        "xdigit" => Some(ch.is_ascii_hexdigit()),
+        _ => None,
+    }
+}
+
+fn match_posix_bracket_class(pat: &[char], ch: char) -> Option<(bool, usize)> {
+    if !matches!((pat.first(), pat.get(1)), (Some(&'['), Some(&':'))) {
+        return None;
+    }
+
+    for idx in 2..pat.len().saturating_sub(1) {
+        if pat[idx] == ':' && pat[idx + 1] == ']' {
+            let name: String = pat[2..idx].iter().collect();
+            return posix_bracket_class_matches(&name, ch).map(|matched| (matched, idx + 2));
+        }
+    }
+
+    None
+}
+
+fn match_bracket_class(pat: &[char], ch: char) -> Option<(bool, usize)> {
+    if pat.first() != Some(&'[') {
+        return None;
+    }
+
+    let mut idx = 1;
+    let negated = matches!(pat.get(idx), Some('!' | '^'));
+    if negated {
+        idx += 1;
+    }
+
+    let class_start = idx;
+    let mut matched = false;
+
+    while idx < pat.len() {
+        if pat[idx] == ']' && idx > class_start {
+            return Some((matched != negated, idx + 1));
+        }
+
+        if let Some((posix_matched, consumed)) = match_posix_bracket_class(&pat[idx..], ch) {
+            matched |= posix_matched;
+            idx += consumed;
+        } else if idx + 2 < pat.len() && pat[idx + 1] == '-' && pat[idx + 2] != ']' {
+            let start = pat[idx];
+            let end = pat[idx + 2];
+            if start <= ch && ch <= end {
+                matched = true;
+            }
+            idx += 3;
+        } else {
+            if pat[idx] == ch {
+                matched = true;
+            }
+            idx += 1;
+        }
+    }
+
+    None
 }
 
 fn glob_match_inner(pat: &[char], text: &[char]) -> bool {
@@ -91,14 +192,22 @@ fn glob_match_inner(pat: &[char], text: &[char]) -> bool {
         return text.is_empty();
     }
 
-    // Handle ** (matches any chars including /)
+    // Handle ** (matches any chars, with **/ advancing on path boundaries)
     if pat.len() >= 2 && pat[0] == '*' && pat[1] == '*' {
-        let rest = if pat.len() > 2 && pat[2] == '/' {
-            &pat[3..]
-        } else {
-            &pat[2..]
-        };
-        // Try matching ** against 0..n chars
+        if pat.len() > 2 && pat[2] == '/' {
+            let rest = &pat[3..];
+            if glob_match_inner(rest, text) {
+                return true;
+            }
+            for i in 0..text.len() {
+                if text[i] == '/' && glob_match_inner(rest, &text[i + 1..]) {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        let rest = &pat[2..];
         for i in 0..=text.len() {
             if glob_match_inner(rest, &text[i..]) {
                 return true;
@@ -107,12 +216,9 @@ fn glob_match_inner(pat: &[char], text: &[char]) -> bool {
         return false;
     }
 
-    // Handle * (matches any chars except /)
+    // Handle * (matches any chars)
     if pat[0] == '*' {
         for i in 0..=text.len() {
-            if i > 0 && text[i - 1] == '/' {
-                break;
-            }
             if glob_match_inner(&pat[1..], &text[i..]) {
                 return true;
             }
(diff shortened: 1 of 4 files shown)


crates/sem-cli/src/commands/diff.rs
```diff
@@ -78,27 +78,123 @@ fn split_on_separator(args: Vec<String>) -> (Vec<String>, Vec<String>) {
...
+
+    git.as_ref().and_then(|git| git.as_ref())
+}
+
+fn is_valid_git_rev(git: &mut Option<Option<GitBridge>>, cwd: &str, arg: &str) -> bool {
+    get_git_bridge(git, cwd).is_some_and(|git| git.is_valid_rev(arg))
+}
+
+/// Simple glob matching for pathspecs. Supports `*`, `**`, `?`, and bracket classes.
 fn glob_match(pattern: &str, path: &str) -> bool {
```
rs545837, line 107: `glob_match` converts pattern and path to `Vec<char>` on every call. For the common case of ASCII-only paths (vast majority of file paths), this could work directly on byte slices (`&[u8]`) to avoid two heap allocations per match. Could keep the char-based path as a fallback for non-ASCII, or just operate on bytes since glob matching doesn't need Unicode awareness beyond literal comparison.

crates/sem-cli/src/commands/diff.rs
```diff
@@ -107,38 +203,53 @@ fn glob_match_inner(pat: &[char], text: &[char]) -> bool {
...
+        {
+            return class_matches && glob_match_inner(&pat[consumed..], &text[1..]);
+        }
     }
 
     // Literal character
     !text.is_empty() && pat[0] == text[0] && glob_match_inner(&pat[1..], &text[1..])
 }
 
+fn literal_pathspec_matches(file_path: &str, spec: &str) -> bool {
```
rs545837, line 378: `literal_pathspec_matches` uses `format!("{dir}/")` and `format!("{spec}/")` for prefix checks — two allocations per call. Could avoid these with:

```rust
file_path.starts_with(spec) && file_path.as_bytes().get(spec.len()) == Some(&b'/')
```

Minor, but this is called per file per pathspec during filtering.

crates/sem-cli/src/commands/diff.rs
```diff
@@ -425,19 +578,145 @@ fn is_git_binary_payload_line(line: &str) -> bool {
         && !line.starts_with("+++ ")
 }
 
+fn parse_git_path_token(input: &str) -> Option<(String, &str)> {
```
rs545837: `parse_git_path_token` always returns an owned `String`, but callers like `parse_diff_git_paths` and `is_binary_files_marker` often just compare the result against other strings. Returning `Cow<'_, str>` for the unquoted branch would avoid the allocation in the common case where the path doesn't need unescaping:

```rust
fn parse_git_path_token(input: &str) -> Option<(Cow<'_, str>, &str)>
```

rs545837 (commented):
Performance review — a few allocation-related optimizations.

crates/sem-cli/src/commands/diff.rs
Iron-Ham: Addressed in 63fe380. `glob_match` now takes an ASCII fast path that operates on byte slices and avoids building `Vec<char>` for the common path. Non-ASCII still falls back to the existing char matcher so behavior stays conservative.

crates/sem-cli/src/commands/diff.rs
Iron-Ham: Addressed in 63fe380. The literal pathspec prefix checks now use direct prefix + byte-boundary checks instead of allocating `format!("{dir}/")` / `format!("{spec}/")` strings per match.

crates/sem-cli/src/commands/diff.rs
Iron-Ham: Addressed in 63fe380. `parse_git_path_token` now returns `Cow<_, str>` so unquoted paths are borrowed in the common case, while quoted/escaped paths still allocate after unescaping. I updated the call sites including diff-git path parsing and binary marker handling.

rs545837:

PR #307, opened 2026-06-01
Consolidate cache and verify behavior

### LINKED ISSUE #249: Incremental cache drops a clean caller's edge into a content-clean entity in an edited file

### LINKED ISSUE #250: sem verify always uses the cache and has no --no-cache, producing CI false negatives

### LINKED ISSUE #251: Cache directory .sem is written into the user's working tree and is not gitignored

Iron-Ham:
Summary
- Consolidates cache relocation, verify cache controls, and incremental cache edge preservation.
- Keeps CLI and MCP cache behavior aligned while preserving stale-target graph edges.

Supersedes #278, #293, #301.
Closes https://github.com/Ataraxy-Labs/sem/issues/251
Fixes https://github.com/Ataraxy-Labs/sem/issues/250
Fixes https://github.com/Ataraxy-Labs/sem/issues/249

Test plan
- cargo test -p sem-cli cache
- cargo test -p sem-cli --test verify_cli
- cargo test -p sem-cli --test cache_location
- cargo test -p sem-mcp cache
- cargo test -p sem-core incremental

### DIFF
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -21,8 +21,9 @@ pub struct DiskCache {
 
 impl DiskCache {
     pub fn open(repo_root: &Path) -> Result<Self, rusqlite::Error> {
-        let cache_dir = repo_root.join(".sem");
-        std::fs::create_dir_all(&cache_dir).ok();
+        let cache_dir = shared_cache::cache_dir_for_repo(repo_root)
+            .ok_or_else(|| rusqlite::Error::InvalidPath(repo_root.to_path_buf()))?;
+        shared_cache::create_cache_dir(&cache_dir)?;
         let db_path = cache_dir.join("cache.db");
         let conn = Connection::open(db_path)?;
 
@@ -373,26 +374,26 @@ impl DiskCache {
         })
     }
 
-    /// Incrementally update the cache: only rewrite stale file entries.
-    pub fn save_incremental(
+    /// Incrementally update the cache with graph-repair metadata.
+    pub fn save_incremental_with_repair_metadata(
         &self,
         root: &Path,
         all_files: &[String],
         stale_files: &[String],
         graph: &EntityGraph,
         entities: &[SemanticEntity],
+        repair_changed_clean_entity_ids: bool,
     ) -> Result<(), rusqlite::Error> {
         let source_stale_files: Vec<&String> = stale_files
             .iter()
             .filter(|file| !shared_cache::is_manifest_file_name(file))
             .collect();
-        let stale_set: HashSet<&str> = source_stale_files
+        let source_stale_set: HashSet<&str> = source_stale_files
             .iter()
             .map(|file| file.as_str())
             .collect();
 
         let tx = self.conn.unchecked_transaction()?;
-        let mut deleted_cached_files = Vec::new();
 
         // Delete stale file entries
         {
@@ -402,26 +403,32 @@ impl DiskCache {
             }
         }
 
-        // Delete files that are no longer in the file list (deleted from disk)
-        {
-            let current_set: HashSet<&str> = all_files
-                .iter()
-                .map(|s| s.as_str())
-                .filter(|path| !shared_cache::is_manifest_file_name(path))
-                .collect();
+        let current_set: HashSet<&str> = all_files
+            .iter()
+            .map(|s| s.as_str())
+            .filter(|path| !shared_cache::is_manifest_file_name(path))
+            .collect();
+        let cached_paths: Vec<String> = {
             let mut cached_stmt = tx.prepare("SELECT path FROM files")?;
-            let cached_paths: Vec<String> = cached_stmt
+            cached_stmt
                 .query_map([], |row| row.get(0))
                 .map(|rows| rows.filter_map(|r| r.ok()).collect())
-                .unwrap_or_default();
-            let mut del_files = tx.prepare("DELETE FROM files WHERE path = ?1")?;
-            for path in &cached_paths {
-                if !shared_cache::is_cache_manifest_key(path)
+                .unwrap_or_default()
+        };
+        let deleted_cached_files: Vec<String> = cached_paths
+            .into_iter()
+            .filter(|path| {
+                !shared_cache::is_cache_manifest_key(path)
+                    && !shared_cache::is_manifest_file_name(path)
                     && !current_set.contains(path.as_str())
-                {
-                    del_files.execute(params![path])?;
-                    deleted_cached_files.push(path.clone());
-                }
+            })
+            .collect();
+
+        // Delete files that are no longer in the file list (deleted from disk)
+        {
+            let mut del_files = tx.prepare("DELETE FROM files WHERE path = ?1")?;
+            for path in &deleted_cached_files {
+                del_files.execute(params![path])?;
             }
         }
 
@@ -440,8 +447,9 @@ impl DiskCache {
 
         shared_cache::refresh_manifest_entries(&tx, root)?;
 
-        // Delete entities for stale files
-        {
+        if repair_changed_clean_entity_ids {
+            tx.execute("DELETE FROM entities", [])?;
+        } else {
             let mut del = tx.prepare("DELETE FROM entities WHERE file_path = ?1")?;
             for f in &source_stale_files {
                 del.execute(params![f])?;
@@ -451,31 +459,34 @@ impl DiskCache {
             }
         }
 
-        // Insert new entities for stale files
         {
             let mut ins = tx.prepare(
                 "INSERT OR REPLACE INTO entities (id, name, entity_type, file_path, start_line, end_line, content, content_hash, structural_hash, parent_id, metadata_json) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11)",
             )?;
             for e in entities {
-                if stale_set.contains(e.file_path.as_str()) {
-                    let metadata_json = e
-                        .metadata
-                        .as_ref()
-                        .and_then(|m| serde_json::to_string(m).ok());
-                    ins.execute(params![
-                        e.id,
-                        e.name,
-                        e.entity_type,
-                        e.file_path,
-                        e.start_line as i64,
-                        e.end_line as i64,
-                        e.content,
-                        e.content_hash,
-                        e.structural_hash,
-                        e.parent_id,
-                        metadata_json,
-                    ])?;
+                if !repair_changed_clean_entity_ids
+                    && !source_stale_set.contains(e.file_path.as_str())
+                {
+                    continue;
                 }
+
+                let metadata_json = e
+                    .metadata
+                    .as_ref()
+                    .and_then(|m| serde_json::to_string(m).ok());
+                ins.execute(params![
+                    e.id,
+                    e.name,
+                    e.entity_type,
+                    e.file_path,
+                    e.start_line as i64,
+                    e.end_line as i64,
+                    e.content,
+                    e.content_hash,
+                    e.structural_hash,
+                    e.parent_id,
+                    metadata_json,
+                ])?;
             }
         }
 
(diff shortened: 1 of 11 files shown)


crates/sem-cli/src/cache.rs
```diff
@@ -440,42 +435,31 @@ impl DiskCache {
...
-            }
-            for f in &deleted_cached_files {
-                del.execute(params![f])?;
-            }
-        }
-
-        // Insert new entities for stale files
+        // Rewrite entity rows as a complete snapshot; graph repair can change
+        // entity IDs in files whose source bytes are still cached.
+        tx.execute("DELETE FROM entities", [])?;
```
rs545837: The change from incremental entity writes (stale files only) to full snapshot (`DELETE FROM entities` + re-insert all) makes sense for correctness when graph repair changes entity IDs in cached files. But for large repos this could be expensive — e.g. 50K entities rewritten on every incremental save.

A middle ground: track whether graph repair actually changed any IDs in non-stale files (a boolean flag from `build_incremental`). If it didn't (the common case), keep the old incremental path. Only fall back to full snapshot when repair actually mutated cached entity IDs. This preserves the correctness fix while avoiding the write amplification in the typical case.

rs545837 (commented):
Performance note on the incremental cache change.

crates/sem-cli/src/cache.rs
Iron-Ham: Addressed in c0a9045. Incremental graph builds now return repair metadata, and the CLI/MCP cache save paths only do the full entity snapshot when graph repair changed IDs in clean cached files. The common path is back to stale-file-only entity rewrites, including deleted-file cleanup. I also kept the old public MCP `save_incremental` signature as a compatibility wrapper and added repair-aware cache tests for both paths.

Iron-Ham:
Pushed a runtime performance follow-up in `3aae082`.

What changed:
- cached/incremental graph rebuilds precompute entity/file and stale/current ID sets instead of repeatedly scanning all entities
- affected clean file paths are tracked as affected IDs are discovered, so scope resolution avoids a file-by-entity search when selecting files to revisit

Validation:
- `cargo test --manifest-path crates/Cargo.toml -p sem-core build_incremental_with_metadata -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core parser::graph::tests`

crates/sem-core/src/parser/graph.rs
```diff
@@ -570,31 +646,208 @@ impl EntityGraph {
...
+                let ext = entity
+                    .file_path
+                    .rfind('.')
+                    .map(|i| &entity.file_path[i..])
+                    .unwrap_or("");
+                if crate::parser::plugins::code::languages::get_language_config(ext).is_none() {
+                    continue;
+                }
+
+                if !text_mentions_any_name(&entity.content, &affected_target_names) {
```
rs545837, line 761: **Performance: incremental rebuild scans ALL clean entities — O(total_words_in_codebase)**

This loop runs `text_mentions_any_name` on every clean entity's content, then `strip_comments_and_strings` (which allocates a full String copy) on those that pass the first check. On a 500K-entity codebase with a 1-file edit, this scans the entire codebase's text content on the incremental path — the path that is specifically designed to be fast for small edits.

With 500K entities averaging ~20 lines each, this is tens of millions of words scanned and hundreds of thousands of String allocations.

Suggestion: use the `symbol_table` (already built above) to find which entities share a name with `affected_target_names`, then only scan entities in the same file as those candidates. This turns the scan from O(all_entities × words_per_entity) to O(candidate_entities). Alternatively, build an inverted index from names → entity IDs once and reuse it.

crates/sem-core/src/parser/graph.rs
```diff
@@ -570,31 +646,208 @@ impl EntityGraph {
...
+                .collect();
+            let new_stale_file_paths: HashSet<&str> = new_stale_entity_ids
+                .iter()
+                .filter_map(|entity_id| entity_file_paths.get(*entity_id).copied())
+                .collect();
+            let clean_file_import_tokens: HashMap<&str, Vec<String>> = all_file_paths
+                .iter()
+                .filter(|file_path| !stale_set.contains(file_path.as_str()))
+                .filter_map(|file_path| {
+                    let Ok(content) = std::fs::read_to_string(root.join(file_path)) else {
```
rs545837: **Performance: re-reads every clean file from disk on the incremental path**

`std::fs::read_to_string(root.join(file_path))` runs for every non-stale file, then `content_import_tokens_for_file` applies multiple compiled regexes per stale file path. With 5000 clean files and 10 stale file paths, that's 5000 disk reads + 50,000 regex applications on the incremental path.

The file content is already available — entities store their content, and the files were already read during the initial `build()`. Re-reading from disk is redundant I/O that dominates incremental rebuild time.

Suggestion: reuse already-loaded content (entity content strings or cached file content from the initial build) instead of re-reading from disk. For import detection, only the first ~20 lines of a file typically matter — consider scanning just file-level entities' content or capping the scan range.

PR #308, opened 2026-06-01
Consolidate graph and scope resolution

### LINKED ISSUE #195: graph container inherits child call edges
## Summary

`sem graph` (and downstream `sem impact`, `sem context`) emits a phantom call edge from a container entity (class, struct, namespace) to every external symbol referenced by its child methods. The container does not literally call the symbol — only its method body does — but the edge appears under the container too.

This systematically over-counts container-level dependencies. `sem impact ClassName` reports every callee its methods reach, when the correct set is "methods of this class plus their dependents."

## Severity

P1 — wrong-edges correctness bug in graph traversal. Affects every impact/context query on a class-shaped entity. Verified on a release-build binary.

## Environment

## Reproduction

```bash
tmp=$(mktemp -d) && cd "$tmp"
git init -q

cat > app.ts <<'EOF'
export function helper() { return 1; }
export class Service {
  method() { return helper(); }
}
EOF

sem graph --json --no-cache --file-exts .ts | jq -c '.edges[]'
```

## Observed

```json
{"fromEntity":"app.ts::class::Service",         "refType":"calls","toEntity":"app.ts::function::helper"}
{"fromEntity":"app.ts::class::Service::method", "refType":"calls","toEntity":"app.ts::function::helper"}
```

Two edges. Only one is real.

## Expected

A single edge:

```json
{"fromEntity":"app.ts::class::Service::method","refType":"calls","toEntity":"app.ts::function::helper"}
```

The class entity itself contains no call expression to `helper` — only its `method` body does.

## Code references

Reference extraction in [crates/sem-core/src/parser/graph.rs](https://github.com/Ataraxy-Labs/sem/blob/2447cf74f7ac35466ac613b5a7401b1bf87431de/crates/sem-core/src/parser/graph.rs) buckets references by line range, so any call inside the class span is attributed to both the class and the containing method. Parent-child filtering excludes references that hit a child entity, but external calls (`helper` lives in the same file but is not nested inside the class) are not filtered.

## Suggested fix

When emitting a reference, find the *innermost* enclosing entity (by line range) and emit the edge only from that entity. Reference attribution should be unique per call site, not duplicated up the parent chain.

## Independently confirmed

Independently surfaced by a parallel audit; reproduced on release binary.

### LINKED ISSUE #248: Bag-of-words fallback invents call edges the scope resolver already declined
**Severity:** High
**Component:** graph / dependency-edge resolution (Phase-2 bag-of-words fallback)
**Affected commands:** `sem graph`, `sem impact`, `sem context`, `sem verify`

## Summary

After the precise scope resolver runs, a cruder Phase-2 "bag-of-words" pass tokenizes each entity's content and links any token that name-matches a symbol in the table. This pass re-resolves names the scope resolver had already (correctly) declined to link, inventing false call edges. Two everyday patterns trigger it: a local variable that shadows a function name, and a method call on a receiver whose type cannot be resolved (e.g. an external-crate type). The false edges pollute `sem impact` blast-radius and produce false `sem verify` arity mismatches.

### Finding A — local variable shadowing a function name

```sh
cat > b.py <<'EOF'
def total(items):
    return sum(items)

def report():
    total = 0
    for i in range(10):
        total = total + i
    return total
EOF

sem graph b.py --json          # expect no edge report -> total
sem impact total --file b.py --json
```

### Finding B — method call on an unresolved external-typed receiver

```sh
cat > lib.rs <<'EOF'
struct Thing;

impl Thing {
    // an associated fn named `build` taking three params
    fn build(a: i32, b: i32, c: i32) -> i32 {
        a + b + c
    }
}

struct Wrapper {
    inner: Vec<i32>,
}

impl Wrapper {
    // `self.inner.iter()` returns an external (std) iterator type;
    // `.build` here would be on an unresolved receiver type
    fn run(&self) {
        let builder = some_external_crate_value();
        let _ = builder.build();
    }
}
EOF
sem graph lib.rs --json
```

Finding B reproduces directly on the `sem` codebase itself: `sem verify` on a clean checkout reports two arity mismatches at `crates/sem-mcp/src/server.rs:164` and `:205`, both of which are `builder.build()` where `builder` has type `ignore::WalkBuilder` (an external crate type). Those resolve by name to the only local `build` symbol, `EntityGraph::build` (3 params).

### Finding A

`sem graph b.py --json` emits a false edge `report -> total`. `sem impact total --file b.py --json` lists `report` as a dependent, even though `report` only uses a local variable named `total` and never calls the function `total`. The edge is additionally mislabeled `typeref` rather than `calls`.

### Finding B (on the `sem` repo)

```
sem verify --json
[
  {"caller": "SemServer", "callee": "build", "actual_args": 0, "expected_min": 3, "expected_max": 3, "file": "crates/sem-mcp/src/server.rs", "line": 164},
  {"caller": "SemServer", "callee": "build", "actual_args": 0, "expected_min": 3, "expected_max": 3, "file": "crates/sem-mcp/src/server.rs", "line": 205}
]
```

`sem graph --json` contains a false `calls` edge from the impl-block container `server.rs::impl::SemServer@L70` to `graph.rs::impl::EntityGraph::build`. `sem impact build --file crates/sem-core/src/parser/graph.rs` reports `SemServer` as an affected dependent. (The other `server.rs` dependent, `get_or_build_graph`, is legitimate: it really does call `EntityGraph::build(repo_root, file_paths, &self.registry)` with three arguments at `crates/sem-mcp/src/server.rs:360`, which is why it does not appear in the `verify` mismatch list.)

## Expected

No edge `report -> total` (Finding A); no edge from any `server.rs` entity to `EntityGraph::build` and no `verify` arity mismatch for `builder.build()` (Finding B). A name the scope resolver bound as a local variable, or a method call on a receiver whose type cannot be resolved to the owner of the candidate definition, must not produce a call edge.

## Root cause

The scope resolver does the right thing in both cases — it treats `total` as a local binding (no edge), and `resolve_ref`'s `MethodCall` arm in `crates/sem-core/src/parser/scope_resolve.rs:3144-3266` returns `None` when the receiver type is unresolved. But the Phase-2 bag-of-words pass overrides that conclusion.

The fallback runs in `crates/sem-core/src/parser/graph.rs:374-432` (and is mirrored in `build_incremental` around `:734-813`). It tokenizes each entity's stripped content via `extract_references_with_stripped`, skips any token in `consumed_words`, and otherwise resolves the token against the same-file symbol table (`graph.rs:405-431`), emitting an edge. The `consumed_words` set is assembled in `crates/sem-core/src/parser/graph.rs:35-60` (`build_scope_consumed_words` / `add_scope_reference_words`) and populated per entity in the dot-chain phase at `crates/sem-core/src/parser/graph.rs:299-372`. Local bindings the scope resolver declined are not added to `consumed_words`, so the token survives to the symbol-table match.

Finding B is attributed to the enclosing impl-block container (`SemServer@L70`) because `consumed_words` is keyed per inner-method entity, while the container's content text spans the method bodies and is tokenized separately. `infer_ref_type` also mislabels the lowercase token as a `typeref` edge in Finding A.

## Suggested fix

Have the scope resolver record names it bound as local variables (and the method member of any unresolved-receiver method call) into the per-entity `consumed_words` so the Phase-2 pass cannot re-resolve them. Equivalently, do not emit a call edge from the bag-of-words pass for a token that the scope resolver already saw and declined.

## Related issues

#195 and PR #245 ("container entities inheriting child call edges") address the container-level case — including the false `SemServer` impact dependent in Finding B, which is attributed to the impl-block container. The case here is at the method / owning-entity level: the Phase-2 bag-of-words fallback re-resolves a name the scope resolver already bound and attributes the false edge to the innermost owning entity itself (in Finding A, `report` is that owning function), which the "attribute to innermost owner" approach does not remove. Related to #148 (closed) and #123 (closed, "sem verify produces too many false positives").

### LINKED ISSUE #252: Graph build scales super-linearly (~cubically) with entity nesting depth
**Severity:** Medium
**Component:** graph / scope resolution (graph build, not extraction)
**Affected commands:** `sem graph`, `sem impact`, `sem context`

## Summary

Building the dependency graph for a deeply-nested file scales roughly cubically with nesting depth, while `sem entities` on the same file stays fast — isolating the cost to graph construction and scope resolution rather than parsing or entity extraction. Generated, macro-expanded, or hostile input with deep nesting can stall `sem graph` (and therefore `sem impact` and `sem context`, which build the same graph) for minutes.

## Reproduction

```sh
gen() {
  python3 - "$1" <<'PY'
import sys
n = int(sys.argv[1]); s = "class L0 {\n"
for i in range(1, n): s += "  " * i + f"L{i} = class {{\n"
s += "  " * n + "method() { return 1; }\n"
for i in range(n - 1, 0, -1): s += "  " * i + "};\n"
s += "}\n"; open("deep.ts", "w").write(s)
PY
}

for depth in 100 200 400 600 800; do
  gen "$depth"; git add -A && git commit -qm init >/dev/null 2>&1
  s=$(python3 -c "import time;print(time.time())")
  sem graph deep.ts >/dev/null 2>&1
  e=$(python3 -c "import time;print(time.time())")
  printf 'graph depth=%-4s secs=%.2f\n' "$depth" "$(python3 -c "print($e-$s)")"
  cd - >/dev/null
done
```

## Observed

Graph-build time grows far faster than linearly in depth (release build):

| depth | `sem graph` | `sem entities` |
|------:|------------:|---------------:|
| 100   | 0.38s       | —              |
| 200   | 1.34s       | —              |
| 400   | 10.1s       | —              |
| 600   | 29.1s       | —              |
| 800   | 107s        | 1.17s          |

Doubling the depth multiplies graph-build time by roughly 7.5–10.6×, consistent with a cubic trend. By contrast, `sem entities deep.ts` at depth 800 completes in 1.17s, so parsing and entity extraction are effectively linear; the blowup is in graph build / scope resolution. `sem impact` and `sem context` inherit the cost because they build the same graph.

## Expected

Graph construction should scale close to linearly (or at worst mildly super-linearly) with nesting depth, in line with extraction.

## Root cause

The cost is in `EntityGraph::build` (`crates/sem-core/src/parser/graph.rs:138-477`) and the scope machinery it drives: `build_scopes_from_ast` (`crates/sem-core/src/parser/scope_resolve.rs:499`) and the per-entity reference resolution loop, whose scope-chain walks climb the full ancestor chain for every reference (`lookup_scope_chain`, `is_local_binding_in_scopes`, `lookup_type_in_scopes` at `crates/sem-core/src/parser/scope_resolve.rs:3292-3343`). With deep nesting, the per-entity range filtering and repeated scope-chain ascents compound into ~O(entities²) (or worse) overall work. `sem entities` does not run this path, which is why it stays fast on the identical input.

## Suggested fix

Index entity ranges so reference resolution does not re-scan all entities per reference (avoid the O(entities²) range filtering), and cap scope-resolution nesting depth (or memoize scope-chain lookups) so a pathologically deep file degrades gracefully instead of stalling.

## Related issues

Distinct from but thematically related to a separate issue covering the Rust structural-hash quadratic at extraction time (that one surfaces in `sem entities`; this one is graph-build / scope-resolution and surfaces in `sem graph`). Related to #103 / #104 (closed, stack overflow on nested code), #191 (closed, slow on large monorepos), #105 / #111 (closed), and #112 (merged, index lookups + parallelize) — the super-linear scope-resolution path was not addressed by that work.

### LINKED ISSUE #255: Swift extension members are not resolved as members of the extended type
**Severity:** High
**Component:** parser / Swift scope resolution, graph
**Affected commands:** `sem graph`, `sem impact`, `sem context`, `sem verify`

## Summary

When a Swift method is declared in an `extension`, calls into that method through an instance receiver produce no `calls` edge in the dependency graph, so `sem impact` reports zero dependents for the extension method. Extensions are a pervasive Swift idiom (protocol conformances, organizing members across files), so this silently blinds blast-radius analysis for a large fraction of real method calls.

## Reproduction

```sh
cat > Example.swift <<'EOF'
struct Widget {
    let name: String
}

extension Widget {
    func render() -> String { return name }
}

func draw(widget: Widget) {
    print(widget.render())
}
EOF
sem graph --json --file-exts .swift --no-cache
sem impact render --no-cache --json
```

## Observed

The only edge from `draw` is a `typeref` to the struct `Widget`. There is no `calls` edge from `draw` to `render`, and `sem impact render` returns no dependents.

```
# graph edges
draw --typeref--> Widget        ← only edge; the call to render() is invisible

# impact render
dependents: []
```

## Expected

`draw(widget:)` calls `widget.render()`, so the graph should contain `draw --calls--> Widget::render`, and `sem impact render` should list `draw` as a dependent.

## Root cause

The scope resolver populates its `class_members` map only for entities whose parent's `entity_type` is one of `class | struct | interface | impl | enum | protocol_declaration | object_declaration | companion_object` (`crates/sem-core/src/parser/scope_resolve.rs:143-148`). `extension` is absent from that set, so members declared in an extension are never inserted into `class_members`. Receiver resolution for `receiver.method()` looks up the receiver's type and then `class_members.get(type)` (`scope_resolve.rs:3201-3208`); since `Widget`'s extension members were never indexed, `render` is not found and no edge is emitted.

## Suggested fix

Add `"extension"` to the parent-type match at `scope_resolve.rs:143-148`, and merge an extension's members into the `class_members` entry keyed by the extended type's name (an `extension Widget` contributes its members to `class_members["Widget"]`). This makes `widget.render()` resolve through the existing receiver-type path.

Note that intra-extension calls (a method in the extension calling another member of the same extension) do resolve as a typeref edge; only cross-receiver calls into the extension fail.

### LINKED ISSUE #256: Swift bare instance-property method calls are not resolved
**Severity:** High
**Component:** parser / Swift scope resolution, graph
**Affected commands:** `sem graph`, `sem impact`, `sem context`, `sem verify`

## Summary

A method call on a bare stored-property receiver (`conn.execute(...)` inside an instance method, where `conn` is a stored property of the enclosing type) produces no `calls` edge. The same call written `self.conn.execute(...)` resolves, but idiomatic Swift omits `self`, so the common case is missed and `sem impact` returns an empty blast radius for the callee.

## Reproduction

```sh
cat > Example.swift <<'EOF'
class Connection {
    func execute(query: String) {}
    func commit() {}
}

class Transaction {
    let conn: Connection
    init(conn: Connection) { self.conn = conn }
    func execute(query: String) {
        conn.execute(query: query)
    }
    func commit() {
        conn.commit()
    }
}
EOF
sem graph --json --file-exts .swift --no-cache
sem impact --entity-id 'Example.swift::class::Connection::execute' --json --no-cache
```

## Observed

No edge `Transaction::execute --calls--> Connection::execute` (and likewise for `commit`) is produced, and `sem impact` on `Connection::execute` returns no dependents.

```
# impact Connection::execute
dependents: []
impact: { total: 0 }
```

The repository's own benchmark already records these exact misses. Running the Swift scope benchmark prints:

```
MISSED [swift]: Transaction::execute -> execute (Transaction.execute calls conn.execute)
MISSED [swift]: Transaction::commit -> commit (Transaction.commit calls conn.commit)
```

## Expected

`Transaction.execute` calls `conn.execute(...)` where `conn: Connection`, so the graph should contain `Transaction::execute --calls--> Connection::execute` (and `Transaction::commit --calls--> Connection::commit`), and `sem impact Connection::execute` should list `Transaction::execute` as a dependent.

## Root cause

The `MethodCall` resolver handles `self.attr.method()` (`crates/sem-core/src/parser/scope_resolve.rs:3167-3181`) and chained `var.field.method()` (`scope_resolve.rs:3184-3198`), and resolves a bare `receiver.method()` only by looking up the receiver in the local scope chain via `lookup_type_in_scopes` (`scope_resolve.rs:3201-3208`). A bare stored-property receiver such as `conn` is not a local binding, so `lookup_type_in_scopes` returns `None` and resolution falls through with no edge. There is no branch that, for a bare identifier matching an instance property of the enclosing type, resolves the property's declared type and then its members.

## Suggested fix

When the receiver is a bare identifier that is not a local binding, check whether it names an instance property of the enclosing type; if so, resolve its declared type via `instance_attr_types` (already consulted at `scope_resolve.rs:3173`) and then `class_members` to find the called method. This mirrors the existing `self.conn.method()` path for the no-`self` form.

Note: an `init --calls--> conn` edge may appear alongside this in the graph output; that is a separate artifact of Swift function-body locals being extracted as `property` entities, not part of this finding.

## Related issues

Related to #148 (name-based call resolution); this Swift receiver case was not covered there.

### LINKED ISSUE #257: Swift overloaded function calls all collapse to a single overload
**Severity:** Medium
**Component:** parser / Swift scope resolution, graph
**Affected commands:** `sem graph`, `sem impact`, `sem context`

## Summary

When two Swift functions share a name but differ by argument label (a Swift overload, e.g. `load(id:)` vs `load(name:)`), every call to either name resolves to the same single overload. The call site's argument labels — which uniquely identify the intended overload in Swift — are ignored, so `sem impact` on the overload that no call resolves to returns an empty blast radius.

## Reproduction

```sh
cat > Example.swift <<'EOF'
func load(id: Int) -> String { return "id" }

func load(name: String) -> String { return "name" }

func byId() -> String { return load(id: 1) }

func byName() -> String { return load(name: "x") }
EOF
sem graph --json --file-exts .swift --no-cache
sem impact --entity-id 'Example.swift::function::load@L1' --json --no-cache
```

## Observed

Both `byId` and `byName` resolve to the same overload (the one at L3), and `sem impact` on the other overload (L1) returns no dependents.

```
# graph edges
byId   --calls--> load@L3
byName --calls--> load@L3        ← both collapse to the L3 overload

# impact load@L1
dependents: []
```

## Expected

`byId()` calls `load(id: 1)` and should resolve to the `load(id: Int)` overload; `byName()` calls `load(name: "x")` and should resolve to the `load(name: String)` overload. Each overload should have its correct distinct dependent.

## Root cause

The scope resolver's symbol table is keyed by `entity.name` only (`crates/sem-core/src/parser/scope_resolve.rs:135-139`), collapsing both overloads into one key. Bare-call resolution then takes the first same-file candidate and otherwise falls back to `target_ids.first()` (`scope_resolve.rs:3114-3139`), with no consideration of the call site's argument labels or argument count. Both overloads are in-file, so the resolver returns whichever appears first in the candidate list for every call.

## Suggested fix

For languages with labeled arguments (Swift), disambiguate overloaded candidates by the call site's argument labels (and arity as a secondary signal) before selecting a target, rather than picking the first same-file or `target_ids.first()` candidate.

## Related issues

Related to #140 ("Same-name overloads collapse/disambiguate", fixed by #164/#165); the call-target-selection aspect — picking the correct overload by argument label — was not covered there.

### LINKED ISSUE #258: Swift declaration entities: raw node kinds leak as types, subscript misnamed, deinit and custom operators dropped
**Severity:** Medium
**Affected commands:** `sem entities`, `sem graph`, `sem impact`, `sem context`, `sem mcp`

## Summary

Several Swift declaration forms are mis-modelled at extraction time: some surface their raw tree-sitter node kind as the public entity `type` (and bake it into the entity ID), a `subscript` is named after its first type identifier instead of `subscript`, and a `deinit` block is dropped entirely. These break the type vocabulary consumers rely on, produce misleading entity IDs that agents pass back to `sem impact`/`sem context`, and make a class's teardown invisible.

## Reproduction

```sh
cat > Decls.swift <<'EOF'
protocol Drawable {
    associatedtype Canvas
}

typealias Handler = (Int) -> Void

prefix operator ~~~

class Resource {
    var handle: Int = 0
    init(handle: Int) { self.handle = handle }
    deinit { print("freed") }
    subscript(index: Int) -> Int { return handle + index }
}
EOF
sem entities Decls.swift --json | jq -r '.[] | [.type, .name, (.parent_id // "-")] | @tsv'
```

## Observed

```
protocol_declaration        Drawable    -
associatedtype_declaration  Canvas      Decls.swift::protocol_declaration::Drawable
typealias_declaration       Handler     -
class                       Resource    -
property                    handle      Decls.swift::class::Resource
init_declaration            init        Decls.swift::class::Resource
subscript_declaration       Int         Decls.swift::class::Resource
```

Neither a `deinit` entity nor an entity for the `prefix operator ~~~` declaration appears at all. Note the leaked type strings are baked into the child entity IDs (e.g. `Decls.swift::protocol_declaration::Drawable`).

### Finding A — raw node kinds leak as entity types

`init_declaration`, `subscript_declaration`, `typealias_declaration`, `associatedtype_declaration`, and `protocol_declaration` surface as the entity `type`, in contrast to `class`/`struct`/`enum`/`actor`/`extension`/`property`/`function`, which are clean.

### Finding B — subscript named after its type

`subscript(index: Int) -> Int` is emitted with name `Int` (ID `…::subscript_declaration::Int`), not anything subscript-related.

### Finding C — nameless declarations (`deinit`, custom operators) are dropped

`deinit { print("freed") }` produces no entity, so a change confined to a `deinit` body is attributed to nothing and the class's teardown is invisible to `sem`. A custom operator declaration such as `prefix operator ~~~` is dropped the same way: it is listed in the Swift `entity_node_types` but is never emitted. (An operator *function* such as `static func + (...)` is a `function_declaration` and is extracted correctly as a `function`; only the bare operator *declaration* is lost.)

## Expected

Friendly, stable entity types (`protocol`, `associatedtype`, `type`, `init`/`constructor`, `subscript`); a subscript named `subscript`; a `deinit` entity (Swift allows at most one per type, so the name is unambiguous); and an entity for each custom operator declaration.

## Root cause

`map_node_type` (`crates/sem-core/src/parser/plugins/code/entity_extractor.rs:1545-1595`) has no match arm for `init_declaration`, `deinit_declaration`, `subscript_declaration`, `typealias_declaration`, or `associatedtype_declaration`, so they fall through `other => other` (`entity_extractor.rs:1593`) and the raw kind is used verbatim. Note `type_alias_declaration` and `type_alias` are mapped (`entity_extractor.rs:1557`) but Swift's one-word `typealias_declaration` is not. The `protocol` case is a deliberate but ugly internal convention: `swift_declaration_keyword_type` maps the `protocol` keyword to the string `"protocol_declaration"` (`entity_extractor.rs:748`) so it matches the `class_members` set in `scope_resolve.rs:146`.

For Finding B, `subscript_declaration` has no `name` field, so `extract_name` (`entity_extractor.rs:1174`) falls through to the generic fallback "return the first child whose kind is `identifier` or `type_identifier`" (`entity_extractor.rs:1459-1465`), which returns the parameter/return type `Int`.

For Finding C, `deinit_declaration` likewise has no `name` field and no identifier child, so `extract_name` returns `None` (`entity_extractor.rs:1467`); entity creation is gated on `if let Some(name) = …` (`entity_extractor.rs:387`), so the deinit is dropped with no fallback. `operator_declaration` (e.g. `prefix operator ~~~`) takes the same path — it has no `name` field, so `extract_name` returns `None` and the gated creation at `entity_extractor.rs:387` drops it as well.

## Suggested fix

Add `map_node_type` arms: `init_declaration => "init"` (or `"constructor"`), `deinit_declaration => "deinit"`, `subscript_declaration => "subscript"`, `typealias_declaration => "type"`, `associatedtype_declaration => "associatedtype"`, `operator_declaration => "operator"`. If `protocol_declaration` is renamed to a friendly `protocol`, update the `class_members` match set at `scope_resolve.rs:143-148` in lockstep. In `extract_name`, synthesize stable names for `subscript_declaration` (`subscript`), `deinit_declaration` (`deinit`), and `operator_declaration` (the operator token, e.g. `~~~`) before the generic type-identifier fallback applies.

## Related issues

Related to #153 ("Entity type vocabulary is inconsistent across languages"); these Swift declaration kinds were not covered there.

### LINKED ISSUE #261: Swift scope-resolution benchmark records missed edges but never fails
**Severity:** Low
**Component:** test suite / scope-resolution benchmark
**Affected commands:** `cargo test` (CI), not a runtime command

## Summary

The Swift scope-resolution benchmark test prints known missed edges to stderr but contains no assertion on recall or false negatives, so it passes green even when expected edges are not resolved. Known correctness gaps and future regressions in Swift call resolution cannot be caught by CI through this test.

## Reproduction

```sh
cd crates && cargo test -q -p sem-core --test scope_resolve_bench scope_resolve_swift -- --nocapture
```

## Observed

The test reports two missed expected edges and then passes.

```
MISSED [swift]: Transaction::execute -> execute (Transaction.execute calls conn.execute)
MISSED [swift]: Transaction::commit -> commit (Transaction.commit calls conn.commit)

=== swift Scope Resolution ===
Old (bag-of-words): 20 TP, 2 FN, 0 FP | 91% recall
New (scope-aware):  20 TP, 2 FN, 0 FP | 91% recall
Total edges: old=44, new=35
test result: ok. 1 passed; 0 failed
```

## Expected

A regression suite should fail (or be clearly separated from the pass/fail coverage) when expected Swift edges are missed, so known gaps and future regressions cannot look green in CI.

## Root cause

The test `scope_resolve_swift` (`crates/sem-core/tests/scope_resolve_bench.rs:1144-1148`) delegates to the shared helper `run_scope_resolve_for_lang` (`crates/sem-core/tests/scope_resolve_bench.rs:706-799`). That helper scores true positives and false negatives, prints `MISSED`/summary lines via `eprintln!`, but contains no `assert!` on the result — it only emits an informational note when zero edges are produced. The two missed edges correspond to the expected entries in `get_swift_expected_edges` (`crates/sem-core/tests/scope_resolve_bench.rs:923-924`). Some sibling tests in the same file do assert (for example at `:271`, `:878`, `:885`), but the path exercised by `scope_resolve_swift` does not.

## Suggested fix

Assert in `run_scope_resolve_for_lang` that the false-negative count is zero (or that recall meets a threshold), or move the benchmark out of the default test target and gate CI on it explicitly. The two current Swift misses share the same root cause as bare instance-property method calls not being resolved, so fixing that resolution path also turns this benchmark green.

Iron-Ham:
## Summary
- Consolidates the graph/scope resolution fixes from the parent PR set into one branch.
- Keeps deep scope resolution scaling, container-edge suppression, Swift entity normalization, Swift extension/member resolution, scoped fallback suppression, overload labels, and bare property receiver resolution together.
- Adds local-binding fallback guards so nested/local declarations do not reattach false import or same-file graph edges after scope resolution.

Supersedes #245, #282, #286, #296, #297, #298, #299, and #304.

closes https://github.com/Ataraxy-Labs/sem/issues/195
Fixes https://github.com/Ataraxy-Labs/sem/issues/261
Closes https://github.com/Ataraxy-Labs/sem/issues/248
Fixes https://github.com/Ataraxy-Labs/sem/issues/255
Fixes https://github.com/Ataraxy-Labs/sem/issues/258
Fixes https://github.com/Ataraxy-Labs/sem/issues/257
Closes https://github.com/Ataraxy-Labs/sem/issues/256
Fixes https://github.com/Ataraxy-Labs/sem/issues/252

## Test plan
- `cargo test --manifest-path crates/Cargo.toml -p sem-core scope_resolve_swift`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core scope_resolve`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core graph`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core`

### DIFF
--- a/crates/sem-core/src/parser/graph.rs
+++ b/crates/sem-core/src/parser/graph.rs
@@ -52,6 +52,9 @@ fn build_scope_consumed_words(
 
 fn add_scope_reference_words(words: &mut HashSet<String>, reference: &str) {
     let reference = reference.strip_suffix("()").unwrap_or(reference);
+    let reference = reference
+        .split_once('(')
+        .map_or(reference, |(name, _)| name);
     if let Some((receiver, member)) = reference.rsplit_once('.') {
         if !receiver.is_empty() {
             words.insert(receiver.to_string());
@@ -59,9 +62,236 @@ fn add_scope_reference_words(words: &mut HashSet<String>, reference: &str) {
         if !member.is_empty() {
             words.insert(member.to_string());
         }
+    } else if reference.contains("::") {
+        for part in reference.split("::").filter(|part| !part.is_empty()) {
+            words.insert(part.to_string());
+        }
     } else if !reference.is_empty() {
         words.insert(reference.to_string());
     }
+}
+
+#[derive(Clone, Copy)]
+struct ChildRange<'a> {
+    file_path: &'a str,
+    start_line: usize,
+    end_line: usize,
+    start_byte: Option<usize>,
+    end_byte: Option<usize>,
+}
+
+fn build_child_ranges_by_parent<'a>(
+    entities: &'a [SemanticEntity],
+) -> HashMap<&'a str, Vec<ChildRange<'a>>> {
+    let entity_by_id: HashMap<&str, &SemanticEntity> = entities
+        .iter()
+        .map(|entity| (entity.id.as_str(), entity))
+        .collect();
+    let mut line_starts_by_parent: HashMap<&'a str, Vec<usize>> = HashMap::new();
+    let mut child_ranges_by_parent: HashMap<&'a str, Vec<ChildRange<'a>>> = HashMap::new();
+
+    for child in entities {
+        let Some(parent_id) = child.parent_id.as_deref() else {
+            continue;
+        };
+        let (start_byte, end_byte) = entity_by_id
+            .get(parent_id)
+            .and_then(|parent| {
+                let parent_line_starts = line_starts_by_parent
+                    .entry(parent_id)
+                    .or_insert_with(|| line_start_offsets(&parent.content));
+                child_content_span_in_parent(parent, child, parent_line_starts)
+            })
+            .map_or((None, None), |(start, end)| (Some(start), Some(end)));
+
+        child_ranges_by_parent
+            .entry(parent_id)
+            .or_default()
+            .push(ChildRange {
+                file_path: child.file_path.as_str(),
+                start_line: child.start_line,
+                end_line: child.end_line,
+                start_byte,
+                end_byte,
+            });
+    }
+
+    for child_ranges in child_ranges_by_parent.values_mut() {
+        child_ranges.sort_unstable_by(|left, right| {
+            match (left.start_byte, right.start_byte) {
+                (Some(left_start), Some(right_start)) => left_start.cmp(&right_start),
+                (Some(_), None) => std::cmp::Ordering::Less,
+                (None, Some(_)) => std::cmp::Ordering::Greater,
+                (None, None) => std::cmp::Ordering::Equal,
+            }
+            .then_with(|| left.end_byte.cmp(&right.end_byte))
+            .then_with(|| left.file_path.cmp(right.file_path))
+            .then_with(|| left.start_line.cmp(&right.start_line))
+            .then_with(|| left.end_line.cmp(&right.end_line))
+        });
+    }
+
+    child_ranges_by_parent
+}
+
+fn child_content_span_in_parent(
+    parent: &SemanticEntity,
+    child: &SemanticEntity,
+    parent_line_starts: &[usize],
+) -> Option<(usize, usize)> {
+    if parent.file_path != child.file_path || child.content.is_empty() {
+        return None;
+    }
+
+    let expected_local_line = child.start_line.checked_sub(parent.start_line)? + 1;
+    if let Some(span) = child_content_span_at_expected_line(
+        &parent.content,
+        &child.content,
+        expected_local_line,
+        parent_line_starts,
+    ) {
+        return Some(span);
+    }
+
+    for (offset, _) in parent.content.match_indices(&child.content) {
+        let local_line = line_for_byte(&parent.content, offset);
+        if local_line == expected_local_line {
+            return Some((offset, offset + child.content.len()));
+        }
+    }
+
+    None
+}
+
+fn child_content_span_at_expected_line(
+    parent_content: &str,
+    child_content: &str,
+    expected_local_line: usize,
+    parent_line_starts: &[usize],
+) -> Option<(usize, usize)> {
+    let line_start = *parent_line_starts.get(expected_local_line.checked_sub(1)?)?;
+    if let Some(span) = content_span_at(parent_content, child_content, line_start) {
+        return Some(span);
+    }
+
+    let line_end = parent_line_starts
+        .get(expected_local_line)
+        .copied()
+        .map(|next_line_start| next_line_start.saturating_sub(1))
+        .unwrap_or(parent_content.len());
+    let line = parent_content.get(line_start..line_end)?;
+
+    let trimmed_line_start = line_start + line.len().saturating_sub(line.trim_start().len());
+    if trimmed_line_start != line_start {
+        if let Some(span) = content_span_at(parent_content, child_content, trimmed_line_start) {
+            return Some(span);
+        }
+    }
+
+    let first_child_line = child_content
+        .split_once('\n')
+        .map_or(child_content, |(line, _)| line);
+    if first_child_line.is_empty() {
+        return None;
+    }
+
+    for (candidate_offset, _) in line.match_indices(first_child_line) {
+        if let Some(span) =
+            content_span_at(parent_content, child_content, line_start + candidate_offset)
+        {
+            return Some(span);
+        }
+    }
+
+    None
+}
+
+fn content_span_at(content: &str, needle: &str, start: usize) -> Option<(usize, usize)> {
+    let end = start.checked_add(needle.len())?;
+    (content.get(start..end) == Some(needle)).then_some((start, end))
+}
+
+fn entity_owns_content_span(
+    entity_id: &str,
+    file_path: &str,
+    source_line: usize,
+    local_start_byte: Option<usize>,
+    local_end_byte: Option<usize>,
+    child_ranges_by_parent: &HashMap<&str, Vec<ChildRange<'_>>>,
+) -> bool {
+    let Some(child_ranges) = child_ranges_by_parent.get(entity_id) else {
+        return true;
+    };
+
+    let child_has_source_line = |child: &ChildRange<'_>| {
+        child.file_path == file_path
+            && source_line >= child.start_line
+            && source_line <= child.end_line
+    };
+
+    let first_without_byte = child_ranges.partition_point(|child| child.start_byte.is_some());
+    if let (Some(start), Some(end)) = (local_start_byte, local_end_byte) {
+        let byte_ranges = &child_ranges[..first_without_byte];
+        let possible_end = byte_ranges.partition_point(|child| {
+            child
+                .start_byte
+                .is_some_and(|child_start| child_start < end)
+        });
+        for child in byte_ranges[..possible_end].iter().rev() {
+            let (Some(child_start), Some(child_end)) = (child.start_byte, child.end_byte) else {
+                continue;
+            };
+            if child_end <= start {
+                break;
+            }
+            if start < child_end && end > child_start && child_has_source_line(child) {
+                return false;
+            }
+        }
+    } else if child_ranges.iter().any(child_has_source_line) {
+        return false;
+    }
+
+    !child_ranges[first_without_byte..]
+        .iter()
+        .any(child_has_source_line)
+}
+
+fn source_line_for_entity_content(entity: &SemanticEntity, local_line: usize) -> usize {
+    entity.start_line + local_line.saturating_sub(1)
+}
+
+fn entity_requires_content_span_filter(
+    entity: &SemanticEntity,
+    child_ranges_by_parent: &HashMap<&str, Vec<ChildRange<'_>>>,
+) -> bool {
+    entity.start_line == entity.end_line
+        || child_ranges_by_parent
+            .get(entity.id.as_str())
+            .map_or(false, |children| {
+                children.iter().any(|child| {
+                    child.start_line == child.end_line
+                        || child.start_line == entity.start_line
+                        || child.end_line == entity.end_line
+                })
+            })
+}
+
+fn line_for_byte(content: &str, byte: usize) -> usize {
+    1 + content[..byte]
+        .bytes()
+        .filter(|current| *current == b'\n')
+        .count()
+}
+
+fn line_start_offsets(content: &str) -> Vec<usize> {
+    let mut starts = vec![0];
+    for (idx, byte) in content.bytes().enumerate() {
+        if byte == b'\n' {
+            starts.push(idx + 1);
+        }
+    }
+    starts
 }
 
 /// A reference from one entity to another.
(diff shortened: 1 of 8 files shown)


crates/sem-core/src/parser/graph.rs
```diff
@@ -47,18 +51,147 @@ fn build_scope_consumed_words(
...
+        let local_line = line_for_byte(&parent.content, offset);
+        if local_line == expected_local_line {
+            return Some((offset, offset + child.content.len()));
+        }
+    }
+
+    None
+}
+
+fn entity_owns_content_span(
```
rs545837, line 214: `entity_owns_content_span` does a linear scan of `child_ranges` on every call, and it's invoked per reference word per entity (hot path). If the child ranges were sorted by `start_byte` in `build_child_ranges_by_parent`, this could use binary search for containment checks instead of `.iter().all()`.

crates/sem-core/src/parser/graph.rs
```diff
@@ -47,18 +51,147 @@ fn build_scope_consumed_words(
...
+                end_line: child.end_line,
+                start_byte,
+                end_byte,
+            });
+    }
+
+    child_ranges_by_parent
+}
+
+fn child_content_span_in_parent(
```
rs545837, line 137: `child_content_span_in_parent` uses `match_indices(&child.content)` which does a full linear scan of the parent's content for each child. For a class with 50 methods, that's 50 linear scans of the same string.

Since we already know the child's expected line (`expected_local_line`), we could use the parent's line-start offsets (computed once) to jump directly to the expected byte range and do a direct comparison, instead of searching the whole parent content.

crates/sem-core/src/parser/graph.rs
```diff
@@ -107,6 +240,274 @@ pub struct EntityInfo {
...
+        end_line: usize,
+    ) -> impl Iterator<Item = &LineReferenceIndex> {
+        let start = start_line.saturating_sub(1).min(self.lines.len());
+        let end = end_line.min(self.lines.len()).max(start);
+        self.lines[start..end].iter()
+    }
+}
+
+impl LineReferenceIndex {
+    fn from_stripped_line(line: &str) -> Self {
```
rs545837, line 441: Minor: `from_stripped_line` allocates 4 collections per line (words Vec, seen_words HashSet, call_words HashSet, import_words HashSet). For a 10K-line file that's 40K allocations for the index. The `words` field stores owned `String`s but is only ever borrowed back as `&str`. Not urgent, but if this shows up in profiling, an arena-based approach or storing word boundaries as `(usize, usize)` byte offsets into the stripped content could reduce allocation pressure.

rs545837 (commented):
Performance review — a few optimization opportunities I spotted in the new code.

crates/sem-core/src/parser/graph.rs
```diff
@@ -107,6 +240,274 @@ pub struct EntityInfo {
...
+        .iter()
+        .map(|(file_path, content, _)| (file_path.as_str(), content.as_str()))
+        .collect();
+
+    file_paths
+        .iter()
+        .filter_map(|file_path| {
+            let ext = file_path.rfind('.').map(|i| &file_path[i..]).unwrap_or("");
+            crate::parser::plugins::code::languages::get_language_config(ext)?;
+
```
Iron-Ham, line 497: Addressed in 7b585fb. `build_reference_indexes` now uses the existing `maybe_par_iter!` pattern so per-file reference index construction can run in parallel when the parallel feature is enabled.

crates/sem-core/src/parser/graph.rs
```diff
@@ -107,6 +240,274 @@ pub struct EntityInfo {
...
+            let content = std::fs::read_to_string(root.join(file_path)).ok()?;
+            Some((
+                file_path.clone(),
+                FileReferenceIndex::from_content(&content),
+            ))
+        })
+        .collect()
+}
+
+fn identifier_tokens(line: &str) -> Vec<(&str, usize)> {
```
Iron-Ham: Addressed in 7b585fb. `identifier_tokens` now returns a streaming iterator instead of allocating a `Vec` per line, and the existing index construction code consumes it directly.

crates/sem-core/src/parser/graph.rs
Iron-Ham: Addressed in 7b585fb. Child ranges are sorted during `build_child_ranges_by_parent`, and `entity_owns_content_span` now uses the sorted byte spans to prune the search instead of scanning every child range on each reference word.

crates/sem-core/src/parser/graph.rs
Iron-Ham: Addressed in 7b585fb. `child_content_span_in_parent` now checks the expected line-derived byte offset first using cached line starts, then falls back to the prior search when the direct check does not match.

crates/sem-core/src/parser/graph.rs
Iron-Ham: Partially addressed in 7b585fb via the `identifier_tokens` iterator change, which removes one per-line allocation source. I left the broader arena / word-boundary storage idea out for now since it is a larger representation change and your note called it minor/profiling-dependent.

crates/sem-core/src/parser/scope_resolve.rs
```diff
@@ -101,6 +226,69 @@ pub(crate) struct PreBuiltLookups {
...
+            } else {
+                None
+            }
+        })
+    }
+}
+
+#[derive(Default)]
+struct ScopeLookupCache {
+    defs: HashMap<(usize, String), Option<String>>,
```
rs545837: **Performance: cache key allocates a String on every probe**

The cache keys are `(usize, String)`, so every call to `lookup_scope_chain_cached`, `is_local_binding_in_scopes_cached`, and `lookup_type_in_scopes_cached` does `name.to_string()` even on cache hits. These are the hottest functions in the resolution pipeline — called for every reference in every file. A single `MethodCall` resolution can probe the cache 3-4 times with the same receiver string.

At 500K entities with ~5 refs each, this is millions of unnecessary heap allocations just for cache lookups.

Options (cheapest to most invasive):
1. Compute `name.to_string()` once at the call site and reuse across multiple cache probes in the same `resolve_ref` call
2. Change the key to `(usize, &'a str)` with a lifetime parameter on the cache
3. Intern strings into an arena so the key is a cheap index/pointer

crates/sem-core/src/parser/scope_resolve.rs
```diff
@@ -2832,6 +3581,282 @@ fn extract_python_module_import(
...
+            entity.entity_type.as_str(),
+            "function" | "method" | "init_declaration"
+        )
+}
+
+fn extract_swift_signature_from_entity_content(
+    entity: &SemanticEntity,
+) -> Option<Vec<Option<String>>> {
+    let language = get_language_config(".swift").and_then(|config| (config.get_language)())?;
+    let mut parser = tree_sitter::Parser::new();
```
rs545837, line 3748: **Performance: new tree-sitter Parser created per entity**

`extract_swift_signature_from_entity_content` creates a fresh `tree_sitter::Parser` and calls `set_language` for every entity that needs signature extraction. Parser construction allocates internal tree-sitter state. In a large Swift codebase this could be called for thousands of entities per file.

Fix: hoist the parser creation out of this function and pass `&mut Parser` in from `build_swift_call_signatures`, reusing a single parser instance across all entities. Tree-sitter parsers are designed to be reused — `parse()` resets internal state automatically.

rs545837:

PR #303, opened 2026-06-01
Fix TypeScript import resolution gaps

Iron-Ham:
Fixes #264

## Summary
- Resolve TypeScript/JavaScript relative imports using ordered extension candidates, including `.mts`, `.cts`, `.mjs`, and `.cjs`.
- Resolve default imports to default export declarations and `export { value as default }` specifiers.
- Add entities and graph edges for named re-exports so barrel aliases participate in dependency and impact traversal.

## Test plan
- `cargo test --manifest-path crates/Cargo.toml -p sem-core`
- `cargo build --manifest-path crates/Cargo.toml -p sem-cli`
- Dummy git repo validation under `/tmp/sem-ts-import-validation` covering default imports, default re-export specifiers, ordered extension fallback, aliased/direct barrel re-exports, impact traversal, and exact `.mts/.cts/.mjs/.cjs` imports.

PR #309, opened 2026-06-02
Consolidate TypeScript and JavaScript parser resolution

### LINKED ISSUE #262: TypeScript `abstract class` and its abstract methods are never extracted as entities

### LINKED ISSUE #263: TypeScript `namespace` / `module` declarations are not extracted; members are silently flattened to top level

### LINKED ISSUE #264: TypeScript import resolution gaps: default imports, barrel re-exports, and `.mts`/`.cts` extensions
**Severity:** High
**Component:** parser / TypeScript import and scope resolution, graph
**Affected commands:** `sem graph`, `sem impact`, `sem context`

## Summary

Three related TypeScript import-resolution gaps cause the dependency graph to miss real edges (and, in one case, to resolve to the wrong file). Each produces wrong or empty `sem impact` results, which directly undercuts the flagship "what does this change affect" workflow. They share the import-resolution code path and are reported together.

### Finding A — default imports create no dependency or impact edge

`import build from './lib'` where `lib.ts` does `export default function makeThing()` produces no edge. The local alias `build` is resolved as if it were a *named* export called `build`, which does not exist, so it never maps to the default-exported `makeThing`.

### Finding B — barrel re-exports are invisible

`export { core as publicCore } from './lib'` is never processed. The re-export alias never enters the import table, so a consumer importing `publicCore` from the barrel is not linked to `core`, and the barrel file yields zero entities.

### Finding C — `.mts` / `.cts` are omitted from the resolver's extension list

The TypeScript import resolver passes a fixed extension list `[".ts", ".tsx", ".js", ".jsx"]` that omits `.mts` and `.cts`, even though extraction supports those extensions and the README documents them. When a `.mts`/`.cts` target cannot be built as a candidate path, resolution falls back to a name-only match that can select the wrong file.

### Finding A — default imports

```sh
cat > lib.ts <<'EOF'
export default function makeThing(): string { return "ok"; }
export function namedThing(): string { return "named"; }
EOF
cat > consumer.ts <<'EOF'
import build, { namedThing as aliasNamed } from './lib';
export function useDefault(): string { return build(); }
export function useNamed(): string { return aliasNamed(); }
EOF

sem graph --json --no-cache
sem impact makeThing --file lib.ts --no-cache
sem impact namedThing --file lib.ts --no-cache
```

### Finding B — barrel re-exports

```sh
cat > lib.ts <<'EOF'
export function core(): string { return "core"; }
EOF
cat > barrel.ts <<'EOF'
export { core as publicCore } from './lib';
EOF
cat > consumer.ts <<'EOF'
import { publicCore } from './barrel';
export function usePublicCore(): string { return publicCore(); }
EOF

sem entities barrel.ts
sem graph --json --no-cache
sem impact core --file lib.ts --no-cache
sem impact publicCore --file barrel.ts --no-cache
```

### Finding C — `.mts`/`.cts` extension gap

```sh
mkdir -p deep
cat > deep/config.mts <<'EOF'
export function loadConfig(): string { return "deep"; }
EOF
# Same-named distractor in a different file/location:
cat > config.ts <<'EOF'
export function loadConfig(): string { return "sibling"; }
EOF
cat > app.ts <<'EOF'
import { loadConfig } from './deep/config.mts';
export function start(): string { return loadConfig(); }
EOF

sem graph --json --no-cache
```

## Observed

**Finding A:** the graph contains exactly one call edge, `consumer.ts::function::useNamed -> lib.ts::function::namedThing`. There is no `useDefault -> makeThing` edge. `sem impact makeThing` reports no dependents; the named-import control `sem impact namedThing` correctly lists `useNamed`.

**Finding B:** `sem entities barrel.ts` is empty. The graph has zero edges. `sem impact core` reports no dependents; `sem impact publicCore` prints `error: Entity 'publicCore' not found`.

**Finding C:** the call edge resolves to the wrong file — `app.ts::function::start -> config.ts::function::loadConfig` instead of `deep/config.mts`. Running the identical fixture with a `.ts` target (`./deep/config.ts`) resolves correctly to `deep/config.ts`, which isolates the extension as the cause. The `.cts` variant behaves the same as `.mts`.

## Expected

`useDefault -> makeThing`, `usePublicCore -> core` (through the `publicCore` alias), and `start -> deep/config.mts::loadConfig` are all edges; the corresponding `sem impact` queries list the dependents.

## Root cause

`crates/sem-core/src/parser/scope_resolve.rs`:

- **Finding A:** the default-import branch in `extract_ts_import` (the `clause_child.kind() == "identifier"` arm at L2517–2523) calls `resolve_import_name(name, name, ...)`, binding the local alias to a *named* export of the same name. There is no `export default` tracking that would map the local name to the module's default export.
- **Finding B:** `extract_imports_from_ast` (L2424–2448) handles only `import_statement` (plus the Python/Rust/Go import forms); there is no arm for `export_statement` with a `source` clause. `extract_ts_import` (L2457–2527) walks only `import_clause`, so `export { A as B } from './m'` and `export * from './m'` never enter the import table.
- **Finding C:** the extension list `[".ts", ".tsx", ".js", ".jsx"]` is passed at L2498, L2515, and L2521, with no `.mts`/`.cts`. In `crates/sem-core/src/parser/import_resolution.rs`, `module_candidates` (L95–113) builds candidate paths by appending exactly those extensions, so a `.mts`/`.cts` target path is never generated; `file_stem` (L141–151) also does not strip `.mts`/`.cts`. When the candidate match fails, resolution falls back to a name-only match, which selected the wrong `config.ts` above. Explicit same-directory `.mts` imports happen to resolve only because `"util.mts".ends_with(".ts")` is true, masking the gap.

## Suggested fix

1. **Default imports:** track `export default` during extraction or module indexing; bind a default import's local name to the source module's default-exported entity regardless of the alias.
2. **Re-exports:** add an `export_statement`-with-`source` handler that records re-export aliases (and `export *`) into the import table, preserving alias chains so consumers resolve through barrels.
3. **`.mts`/`.cts`:** add `.mts` and `.cts` (and arguably `.mjs`/`.cjs`) to the extension list at L2498/L2515/L2521, and teach `file_stem` to strip them.
4. Add regression tests: default-vs-named import controls; direct and aliased re-export; and a `.mts` target with a same-named `.ts` distractor.

## Related issues

Builds on #224 ("Fix graph import resolution for paths and aliases") and #196 ("graph relative imports resolve by basename"); these three cases were not covered by that work. Related to #148.

### LINKED ISSUE #265: Bodyless TypeScript function signatures (overload heads and `declare function`) are dropped

### LINKED ISSUE #266: Object-literal method shorthand is not extracted as nested entities

### LINKED ISSUE #267: TypeScript getter and setter accessors are indistinguishable (both emit as `method <name>`)

Iron-Ham:
## Summary
- consolidate the open TypeScript/JavaScript parser, extraction, accessor, namespace, abstract entity, and import-resolution fixes
- keep the combined graph/scope behavior from the parent PRs in one reviewable branch

Supersedes #303, #305, #291, #289, #288, #285.
Fixes #264.
Fixes #262.
Fixes #265.
Closes #267.
Fixes #266.
Fixes #263.

## Test plan
- cargo test --manifest-path crates/Cargo.toml -p sem-core typescript
- cargo test --manifest-path crates/Cargo.toml -p sem-core js_ts
- cargo test --manifest-path crates/Cargo.toml -p sem-core import_resolution
- cargo test --manifest-path crates/Cargo.toml -p sem-core scope_resolve_typescript
- cargo test --manifest-path crates/Cargo.toml -p sem-cli --test accessor_cli
- cargo check --manifest-path crates/Cargo.toml -p sem-core -p sem-cli
- cargo test --manifest-path crates/Cargo.toml -p sem-core graph


crates/sem-core/src/parser/differ.rs
```diff
@@ -339,22 +340,68 @@ fn strip_children_content(
...
-                excluded.insert(i);
+        let search_start = line_starts.get(start_idx).copied().unwrap_or(0);
+        let search_end = line_starts
+            .get(end_idx.saturating_add(1))
+            .copied()
+            .unwrap_or(content.len())
+            .min(content.len());
+
+        if !child.content.is_empty() && search_start <= search_end {
+            if let Some(relative_start) = content[search_start..search_end].find(&child.content) {
```
rs545837: `content[search_start..search_end].find(&child.content)` is a naive substring search — O(n×m) where n is the search window and m is the child content length. For large parent entities with many children (e.g. a 500-line class with 30 methods), this adds up.

Since we already computed `line_starts` and know the child's start line relative to the parent (`start_idx`), we could jump directly to the expected byte offset:

```rust
let expected_offset = line_starts.get(start_idx).copied().unwrap_or(0);
if content[expected_offset..].starts_with(&child.content) {
    excluded_ranges.push((expected_offset, expected_offset + child.content.len()));
    continue;
}
// fall back to find() only if the direct check fails
```

This makes the common case O(m) instead of O(n×m).

rs545837 (commented):
Performance note on the strip_children_content rewrite.

crates/sem-core/src/parser/differ.rs
Iron-Ham: Addressed in 1819d73. `strip_children_content` now first checks the expected line-start offset with `starts_with`, and only falls back to the bounded `.find()` search if that direct match fails.

Iron-Ham:
Pushed a runtime performance follow-up in `c10963f`.

What changed:
- default-export candidate files are sorted once and reused during TS/JS import resolution
- namespace imports lazily build the top-level entity index only when a namespace import needs it
- namespace imports now resolve one module candidate through the shared import-order helper before registering exports, which addresses the reviewer concern about merging exports from multiple module variants
- added coverage for ordered module variants and single-candidate namespace behavior

Validation:
- `cargo test --manifest-path crates/Cargo.toml -p sem-core test_js_ts_namespace_import`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core test_js_ts_default_import`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core import_resolution`

rs545837:

rs545837:

rs545837:
@Iron-Ham — #308 just went in and this has conflicts now. Needs a rebase onto main before we can merge. Thanks!

I have started realizing why weave-crdts is even more of an interesting thing to have because multiple PRs affecting the same files in a codebase make it a little hard to sort out merge conflicts.

Iron-Ham:
> I have started realizing why weave-crdts is even more of an interesting thing to have because multiple PRs affecting the same files in a codebase make it a little hard to sort out merge conflicts.

👀 
Talk to me more about this? AFAIK we can't instruct GitHub to calculate conflicts any way other than the default, but if you're getting at what I think you're getting at it sounds like you're touching on something highly valuable that solves a lot of pain.

FWIW, I think this set of work that's open now would have been better represented as a _stack_ of open PRs but that's not something GitHub supports through fork PRs 🫠

rs545837:
Yeah it was basically an idea I have been chasing with sem and weave is to think of the file system and the search primitives from first principles now. Because most of our analysis is not line by line or file by file at the moment. 

It's basically on top of weave's working and the CRDT layer on top adds proactive prevention, agents claim entities before editing so concurrent edits on the same function are caught before branches even exist, so that as a concept makes the most sense to me at the moment.

PR #310, opened 2026-06-02
Consolidate Swift extraction

### LINKED ISSUE #254: Swift local variables are extracted as `property` entities and corrupt the dependency graph

### LINKED ISSUE #259: Swift conditional-compilation recovery drops the enclosing type when an interpolated string contains a brace

### LINKED ISSUE #260: Swift multi-binding property declaration drops all but the first name

Iron-Ham:
## Summary
- consolidate the Swift interpolation recovery, multi-binding property extraction, and local-property suppression fixes
- keep the Swift extractor and scope-resolution changes together in one reviewable branch

Supersedes #279, #302, #287.
Closes #259.
Fixes #260.
Fixes #254.

## Test plan
- cargo test --manifest-path crates/Cargo.toml -p sem-core swift
- cargo test --manifest-path crates/Cargo.toml -p sem-core scope_resolve_swift
- cargo test --manifest-path crates/Cargo.toml -p sem-core reorder
- cargo check --manifest-path crates/Cargo.toml -p sem-core

### DIFF
--- a/crates/sem-core/src/model/identity.rs
+++ b/crates/sem-core/src/model/identity.rs
@@ -698,7 +698,14 @@ fn detect_reorders(
             if lnds_set.contains(&i) {
                 continue;
             }
-            changes.push(make_change(after_entity, ChangeType::Reordered, Some(before_entity), commit_sha, author, by_id));
+            changes.push(make_change(
+                after_entity,
+                ChangeType::Reordered,
+                Some(before_entity),
+                commit_sha,
+                author,
+                by_id,
+            ));
         }
     }
 }
--- a/crates/sem-core/src/parser/graph.rs
+++ b/crates/sem-core/src/parser/graph.rs
@@ -40,40 +40,6 @@ use crate::parser::import_resolution::{
 use crate::parser::registry::{resolve_go_method_parent_ids, ParserRegistry};
 use crate::parser::scope_resolve;
 
-fn build_scope_consumed_words(
-    resolution_log: &[scope_resolve::ResolutionEntry],
-) -> HashMap<String, HashSet<String>> {
-    let mut consumed_by_entity: HashMap<String, HashSet<String>> = HashMap::new();
-    for entry in resolution_log {
-        let words = consumed_by_entity
-            .entry(entry.from_entity.clone())
-            .or_default();
-        add_scope_reference_words(words, &entry.reference);
-    }
-    consumed_by_entity
-}
-
-fn add_scope_reference_words(words: &mut HashSet<String>, reference: &str) {
-    let reference = reference.strip_suffix("()").unwrap_or(reference);
-    let reference = reference
-        .split_once('(')
-        .map_or(reference, |(name, _)| name);
-    if let Some((receiver, member)) = reference.rsplit_once('.') {
-        if !receiver.is_empty() {
-            words.insert(receiver.to_string());
-        }
-        if !member.is_empty() {
-            words.insert(member.to_string());
-        }
-    } else if reference.contains("::") {
-        for part in reference.split("::").filter(|part| !part.is_empty()) {
-            words.insert(part.to_string());
-        }
-    } else if !reference.is_empty() {
-        words.insert(reference.to_string());
-    }
-}
-
 #[derive(Clone, Copy)]
 struct ChildRange<'a> {
     file_path: &'a str,
@@ -856,9 +822,9 @@ impl EntityGraph {
                 Some(parsed_files),
                 Some(pre_built),
                 Some(&import_table),
+                false,
             );
-            let consumed_words = build_scope_consumed_words(&result.resolution_log);
-            (result.edges, consumed_words)
+            (result.edges, result.consumed_words)
         } else {
             (vec![], HashMap::new())
         };
@@ -1623,9 +1589,9 @@ impl EntityGraph {
                 pre,
                 None,
                 Some(&import_table),
+                false,
             );
-            let consumed_words = build_scope_consumed_words(&result.resolution_log);
-            (result.edges, consumed_words)
+            (result.edges, result.consumed_words)
         } else {
             (vec![], HashMap::new())
         };
--- a/crates/sem-core/src/parser/plugins/code/entity_extractor.rs
+++ b/crates/sem-core/src/parser/plugins/code/entity_extractor.rs
@@ -471,6 +471,35 @@ fn visit_node(
             }
         }
 
+        if node_type == "property_declaration"
+            && config.id == "swift"
+            && config.entity_node_types.contains(&node_type)
+        {
+            let bindings = extract_swift_property_bindings(node, source);
+            if bindings.len() > 1 {
+                let should_skip = should_skip_entity(config, suppression_context, node_type);
+                if !should_skip {
+                    for binding in bindings {
+                        let entity_type = map_entity_type(node, config);
+                        entities.push(SemanticEntity {
+                            id: build_entity_id(file_path, entity_type, &binding.name, parent_id),
+                            file_path: file_path.to_string(),
+                            entity_type: entity_type.to_string(),
+                            name: binding.name,
+                            parent_id: parent_id.map(String::from),
+                            content_hash: content_hash(&binding.content),
+                            structural_hash: Some(binding.structural_hash),
+                            content: binding.content,
+                            start_line: binding.start_line,
+                            end_line: binding.end_line,
+                            metadata: None,
+                        });
+                    }
+                    continue;
+                }
+            }
+        }
+
         // Go grouped declarations: `var ( a = 1; b = 2 )` should produce
         // one entity per spec instead of collapsing to the first (#149).
         if (node_type == "var_declaration"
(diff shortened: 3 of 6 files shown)


Iron-Ham:
Pushed a runtime performance follow-up in `d6014f5`.

What changed:
- the graph build path no longer materializes synthetic `local_binding` resolution-log rows just to derive fallback consumed words
- scope resolution now returns consumed words directly for the graph path, while the public `resolve_with_scopes` wrapper keeps emitting `local_binding` debug rows for compatibility
- added a compatibility test for the public `local_binding` log behavior

Validation:
- `cargo test --manifest-path crates/Cargo.toml -p sem-core public_scope_resolution_keeps_local_binding_log -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core local_binding -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core dot_chain -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core unresolved_method -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core scope_resolve -- --nocapture`

rs545837:

rs545837:

Iron-Ham:
@rs545837 rebasing 🫡

rs545837:
@Iron-Ham — #308 just went in and this has conflicts again. One more rebase needed, sorry!

Iron-Ham:
@rs545837 updated

rs545837:
@Iron-Ham — #309 just merged and this has conflicts again. One last rebase needed and we're done with the whole stack\!

Iron-Ham:
@rs545837 rebased

PR #312, opened 2026-06-02
Consolidate identity matching

### LINKED ISSUE #246: Entity IDs collide when two sibling entities share a start line, silently dropping changes

### LINKED ISSUE #247: classify_match reports "renamed" for entities whose name never changed

Iron-Ham:
## Summary
- consolidate same-name fuzzy match classification with same-line entity ID disambiguation
- keep identity matching, parent ID rewriting, and same-line reorder behavior together in one reviewable branch

Supersedes #276, #295.
Fixes #247.
Fixes #246.

## Test plan
- cargo test -p sem-core model::identity::tests
- cargo test -p sem-core parser::differ::tests
- cargo test -p sem-cli --test diff_file_compare
- cargo test -p sem-core hotspot
- cargo check -p sem-core
- cargo check --manifest-path crates/Cargo.toml -p sem-core -p sem-cli
- git diff --no-ext-diff --check origin/main...HEAD

### DIFF
--- a/crates/sem-core/src/model/identity.rs
+++ b/crates/sem-core/src/model/identity.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 
 use super::change::{ChangeType, SemanticChange};
@@ -42,26 +43,76 @@ pub struct MatchResult {
     pub changes: Vec<SemanticChange>,
 }
 
+type SameFileSignatureKey<'a> = (&'a str, &'a str, &'a str, Option<&'a str>);
+type RenameSignatureKey<'a> = (&'a str, &'a str, Option<&'a str>);
+const SAME_FILE_SIGNATURE_MIN_SIMILARITY: f64 = 0.3;
+
+struct ContentTokens<'a> {
+    token_count: usize,
+    unique_tokens: HashSet<&'a str>,
+}
+
+struct TokenCache<'a> {
+    tokens: Vec<Option<ContentTokens<'a>>>,
+}
+
+impl<'a> TokenCache<'a> {
+    fn new(len: usize) -> Self {
+        Self {
+            tokens: std::iter::repeat_with(|| None).take(len).collect(),
+        }
+    }
+
+    fn get(&mut self, entities: &[&'a SemanticEntity], idx: usize) -> &ContentTokens<'a> {
+        if self.tokens[idx].is_none() {
+            let content: &'a str = entities[idx].content.as_str();
+            self.tokens[idx] = Some(tokenize_content(content));
+        }
+        self.tokens[idx].as_ref().unwrap()
+    }
+}
+
+fn tokenize_content(content: &str) -> ContentTokens<'_> {
+    let mut token_count = 0;
+    let mut unique_tokens = HashSet::new();
+    for token in content.split_whitespace() {
+        token_count += 1;
+        unique_tokens.insert(token);
+    }
+    ContentTokens { token_count, unique_tokens }
+}
+
+fn jaccard_similarity(a: &HashSet<&str>, b: &HashSet<&str>) -> f64 {
+    let intersection_size = a.intersection(b).count();
+    let union_size = a.len() + b.len() - intersection_size;
+    if union_size == 0 {
+        return 0.0;
+    }
+    intersection_size as f64 / union_size as f64
+}
+
+fn default_similarity_from_tokens(a: &ContentTokens<'_>, b: &ContentTokens<'_>) -> f64 {
+    let (min_c, max_c) = if a.token_count < b.token_count {
+        (a.token_count, b.token_count)
+    } else {
+        (b.token_count, a.token_count)
+    };
+    if max_c > 0 && (min_c as f64 / max_c as f64) < 0.6 {
+        return 0.0;
+    }
+    jaccard_similarity(&a.unique_tokens, &b.unique_tokens)
+}
+
 fn classify_match(before: &SemanticEntity, after: &SemanticEntity) -> ChangeType {
     if before.file_path != after.file_path {
         ChangeType::Moved
     } else if before.parent_id != after.parent_id {
         ChangeType::Moved // intra-file scope move (e.g. method moved between classes)
-    } else {
+    } else if before.entity_type != after.entity_type || before.name != after.name {
         ChangeType::Renamed
+    } else {
+        ChangeType::Modified
     }
-}
-
-fn same_signature_across_file_rename(
-    before: &SemanticEntity,
-    after: &SemanticEntity,
-    before_by_id: &HashMap<&str, &SemanticEntity>,
-    after_by_id: &HashMap<&str, &SemanticEntity>,
-) -> bool {
-    before.file_path != after.file_path
-        && before.entity_type == after.entity_type
-        && before.name == after.name
-        && parent_name(before, before_by_id) == parent_name(after, after_by_id)
 }
 
 fn structural_change_between(before: &SemanticEntity, after: &SemanticEntity) -> Option<bool> {
@@ -143,9 +194,9 @@ fn make_change(
 
 /// Entity matching algorithm:
 /// 1. Exact ID match — same entity ID in before/after → modified or unchanged
-/// 2. Content hash match — same hash, different ID → renamed or moved
+/// 2. Content hash match — same hash, different ID → modified, renamed, or moved
 /// 3. Same signature across file rename → moved, even if content changed
-/// 4. Fuzzy similarity — >80% content similarity → probable rename
+/// 4. Fuzzy similarity — >80% content similarity → modified, renamed, or moved
 pub fn match_entities(
     before: &[SemanticEntity],
     after: &[SemanticEntity],
@@ -199,6 +250,8 @@ pub fn match_entities(
         .iter()
         .filter(|e| !matched_after.contains(e.id.as_str()))
         .collect();
+    let mut unmatched_before_tokens = TokenCache::new(unmatched_before.len());
+    let mut unmatched_after_tokens = TokenCache::new(unmatched_after.len());
 
     // Phase 2: Content hash match (rename/move detection)
     let mut before_by_hash: HashMap<&str, Vec<&SemanticEntity>> = HashMap::new();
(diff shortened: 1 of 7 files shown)


rs545837 (commented):
Performance review — a couple of optimization spots in the new identity matching phase.

crates/sem-core/src/model/identity.rs
```diff
@@ -254,7 +269,63 @@ pub fn match_entities(
         }
     }
 
-    // Phase 3: Same logical signature across a file rename.
+    // Phase 3: Same logical signature within a file.
+    // Collision groups can shrink or grow, changing only the disambiguator
+    // portion of an ID. Match those entities before the generic fuzzy pass.
+    let mut same_file_candidates: Vec<(f64, usize, usize, usize)> = Vec::new();
```
rs545837: Phase 3's inner loop is O(n²) over all unmatched entities:

```rust
for after_entity in unmatched_after {
    for before_entity in unmatched_before {
        if !same_signature_within_file(...) { continue; }
```

Phase 4 (fuzzy) already groups by entity type to avoid this. Phase 3 could do the same — build a map keyed by `(file_path, entity_type, name)` from unmatched_before, then look up candidates directly instead of scanning every before entity. Since `same_signature_within_file` requires all three fields to match, this reduces most iterations to O(1) lookups.

crates/sem-core/src/model/identity.rs
```diff
@@ -254,7 +269,63 @@ pub fn match_entities(
...
+                continue;
+            }
+            if !same_signature_within_file(before_entity, after_entity, &before_by_id, &after_by_id) {
+                continue;
+            }
+
+            let score = similarity_fn
+                .map(|f| f(before_entity, after_entity))
+                .unwrap_or_else(|| default_similarity(before_entity, after_entity));
+            same_file_candidates.push((
```
rs545837: Each candidate pair calls `default_similarity` which re-tokenizes both entities from scratch (`content.split_whitespace().collect()` + two HashSets). Phase 4 already pre-computes token sets for this reason. Could pre-compute token sets for the unmatched entities once before this loop and reuse them here, especially since the same entities flow into Phase 4 anyway.

crates/sem-core/src/model/identity.rs
Iron-Ham: Addressed in 4f01061. Phase 3 now builds a lookup keyed by `(file_path, entity_type, name)` for unmatched-before entities and uses that candidate set before the parent/signature check, avoiding the full unmatched-before scan for each unmatched-after entity.

crates/sem-core/src/model/identity.rs
Iron-Ham: Addressed in 4f01061. `match_entities` now precomputes content token sets for unmatched entities and reuses them in the default similarity path. I also updated the production diff path to use that cached default path instead of wrapping the default plugin method as a callback; there are no in-tree plugin overrides of `compute_similarity`.

Iron-Ham:
Pushed a runtime performance follow-up in `519ca6a`.

What changed:
- token sets for identity matching are cached lazily instead of rebuilt on every comparison
- same-file matching uses signature groups before falling back to broader comparison
- across-file rename matching uses cheap signature indexes to reduce candidate fan-out
- duplicate-ID handling now has a fast path before fuzzy matching

Validation:
- identity matching unit tests
- code extraction tests that cover duplicate entity disambiguation

crates/sem-core/src/model/identity.rs
```diff
@@ -254,95 +306,238 @@ pub fn match_entities(
...
+
+        same_file_candidates.sort_by(|a, b| {
+            b.0.partial_cmp(&a.0)
+                .unwrap_or(Ordering::Equal)
+                .then_with(|| a.1.cmp(&b.1))
+                .then_with(|| a.2.cmp(&b.2))
+                .then_with(|| a.3.cmp(&b.3))
+        });
+
+        for (_score, _line_distance, before_idx, after_idx) in same_file_candidates {
```
rs545837: Phase 3 has no minimum similarity threshold here. When two entities share `(file, type, name, parent)` but have different IDs (line shift), they match unconditionally — even with near-zero content similarity.

**Repro (unit test):**

```rust
#[test]
fn test_phase3_matches_completely_different_content_same_signature() {
    let before = vec![
        make_entity_at(
            "a.ts::function::process@L1", "process",
            "function process(req: Request) { return validateInput(req.body); const result = processData(req.params); return formatResponse(result, req.headers); }",
            "a.ts", 1,
        ),
        make_entity_at(
            "a.ts::function::process@L7", "process",
            "function process(socket: WebSocket): void { const conn = establishConnection(socket.url); conn.onMessage(data => parseProtobuf(data)); conn.onClose(() => cleanupResources(conn.id)); }",
            "a.ts", 7,
        ),
    ];
    let after = vec![
        make_entity_at(
            "a.ts::function::process@L1", "process",
            "function process(req: Request) { return validateInput(req.body); const result = processData(req.params); return sendJSON(result); }",
            "a.ts", 1,
        ),
        make_entity_at(
            "a.ts::function::process@L9", "process",
            "function process(file: File): Promise<string> { const buffer = await readFileAsBuffer(file); const hash = computeSHA256(buffer); await uploadToS3(hash, buffer); return generateCDNUrl(hash); }",
            "a.ts", 9,
        ),
    ];

    let result = match_entities(&before, &after, "a.ts", None, None, None);
    // Phase 3 matches process@L7 (WebSocket) -> process@L9 (file upload) as "modified"
    // despite Jaccard similarity of only 0.185. These share zero domain logic.
    let modified = result.changes.iter().filter(|c| c.change_type == ChangeType::Modified).count();
    assert_eq!(modified, 2); // both reported as modified — the upload handler should be added, not modified
}
```

`process@L7` (WebSocket: `establishConnection`, `parseProtobuf`, `cleanupResources`) and `process@L9` (file upload: `readFileAsBuffer`, `computeSHA256`, `uploadToS3`) share Jaccard = 0.185 but Phase 3 matches them as "modified" because there's no floor check.

**Fix — one line:**
```rust
for (_score, _line_distance, before_idx, after_idx) in same_file_candidates {
    if _score < 0.3 { continue; }  // reject clearly unrelated entities
```

0.3 is generous enough to keep "same function, heavily edited" matches while blocking "totally different function, same name". Phase 5 uses 0.8; Phase 3 can be lower since the signature match provides strong signal.

rs545837 (commented):
Phase 3 needs a minimum similarity threshold — it currently matches entities with completely different content as long as they share the same (file, type, name, parent) signature.

crates/sem-core/src/model/identity.rs
Iron-Ham: Addressed in `05b4517`. Phase 3 same-file signature matching now requires a finite similarity score of at least `0.3` before accepting a pair, so same `(file, type, name, parent)` entities with unrelated content fall through as add/delete instead of being forced into a modified match.

I also added the regression coverage from this review shape: `test_same_file_signature_rejects_unrelated_content` verifies the unrelated same-name handler is reported as added/deleted while the related handler remains modified.

Validation:
- `cargo test --manifest-path crates/Cargo.toml -p sem-core test_same_file_signature_rejects_unrelated_content -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core identity -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core parser::differ::tests -- --nocapture`

PR #319, opened 2026-06-05
Improve graph cache and diff performance

### LINKED ISSUE #313: Profile and reduce sem diff patch and formatter allocation overhead
**Severity:** Low
**Component:** diff CLI / patch input / output formatters
**Affected commands:** `sem diff --patch`, `sem diff --json`, `sem diff --markdown`, `sem diff --terminal --verbose`
**Reproduced on:** source inspection plus generated large-diff fixtures

## Summary

The current highest-priority runtime work is in graph/cache paths used by `sem impact`, but the all-PR performance review also found smaller wins in `sem diff` that should be tracked separately.

Potential hotspots:

- `sem diff --patch` resolves each patch entry's blob contents with a separate `git show` subprocess.
- JSON formatting builds a full `serde_json::Value` tree before serializing.
- Markdown and terminal formatters build whole `Vec<String>`/`String` outputs in memory.
- Verbose diff formatting can recompute line-level diff/hunk data per changed entity.

These are not urgent enough to block the current PR stack, but they are worth a focused follow-up for very large diffs.

## Reproduction

A public generated repo can create a large patch:

```sh
tmp=$(mktemp -d)
cd "$tmp"
git init -q
for i in $(seq 1 1000); do
  printf 'export function f%s() { return %s }\n' "$i" "$i" > "f$i.ts"
done
git add . && git commit -qm init
for i in $(seq 1 1000); do
  printf 'export function f%s() { return %s + 1 }\n' "$i" "$i" > "f$i.ts"
done

git diff --binary >/tmp/sem-large.patch
/usr/bin/time -p sem diff --patch --json </tmp/sem-large.patch >/dev/null
/usr/bin/time -p sem diff --patch --markdown --verbose </tmp/sem-large.patch >/dev/null
```

Source inspection:

```sh
rg -n "git show|diff_json_value|serde_json::json|format_markdown|format_terminal|iter_hunks" crates/sem-cli/src crates/sem-core/src
```

## Observed

Patch mode shells out to `git show` per entry, and the output formatters generally materialize the complete formatted output before writing it. For normal PR-sized diffs this is fine; for large generated patches or broad refactors it adds unnecessary subprocess, allocation, and serialization overhead.

## Expected

Large-diff handling should avoid per-file subprocess fan-out and should stream output where possible.

## Root cause

The diff code paths were built for clarity and stable output contracts first. They do not yet batch blob reads or use streaming serializers/formatters.

## Suggested fix

- Replace per-entry `git show` in patch mode with libgit2 blob reads or a single `git cat-file --batch` session.
- Stream JSON with `serde_json::Serializer` instead of building a full `Value` tree for large results.
- Let markdown/plain/terminal formatters write to an `io::Write` sink or a shared output builder with precomputed capacity.
- Cache line-level diff/hunk computation per before/after content pair when verbose output needs it.
- Add a generated large-diff benchmark to guard the path.

## Related issues

- #269, #271, #272, #273, and #274 cover diff correctness/output-contract issues. This ticket is only about performance and memory behavior after those semantics are stable.

### LINKED ISSUE #314: Compact fallback reference indexes used by graph resolution
**Severity:** Medium
**Component:** parser / graph fallback resolution
**Affected commands:** cold or partial-cache `sem impact`, `sem graph`, `sem verify`, MCP graph/impact tools
**Reproduced on:** source inspection of the graph-scope consolidation branch plus generated fixtures

## Summary

The graph-scope consolidation work bounds fallback reference scanning to direct entity ranges and introduces per-file reference indexes. That is a correctness and asymptotic improvement over repeated full-content scans, but the index representation is still allocation-heavy for large files:

- each indexed line stores owned `String` tokens,
- dot-chain pairs are owned `String` tuples,
- call/import words are separate `HashSet<String>` values per line,
- empty or low-reference lines still carry collection overhead,
- each entity query builds its own dedup state over the selected ranges.

This is a good follow-up after the current PR stack lands because the bounded index gives us a clear place to optimize memory layout without reopening the fallback semantics.

## Reproduction

A generated fixture can make this visible without private code:

```sh
tmp=$(mktemp -d)
cd "$tmp"
git init -q
for f in $(seq 1 200); do
  {
    printf 'export class C%s {\n' "$f"
    for i in $(seq 1 500); do
      printf '  m%s() { this.helper%s(); imported%s(); }\n' "$i" "$i" "$i"
    done
    printf '  helper1() {}\n}\n'
  } > "c$f.ts"
done

git add . && git commit -qm init
/usr/bin/time -p sem graph --no-cache --file-exts .ts >/dev/null
```

Source inspection on the graph-scope branch:

```sh
rg -n "struct LineReferenceIndex|struct FileReferenceIndex|from_content|refs_with_types_in_ranges|dot_chains_in_ranges" crates/sem-core/src/parser/graph.rs
```

## Observed

The fallback index stores owned strings and multiple collections per line, then performs per-query deduplication for every entity range. That is much cheaper than repeatedly parsing full entity content, but it can still become a large allocation and memory-bandwidth cost in generated or large TS/JS files.

## Expected

Fallback indexing should keep the same bounded semantics while using a compact representation proportional to actual references rather than source-line count and repeated owned strings.

## Root cause

The first bounded-reference index favors implementation clarity. It does not intern tokens, compress line entries, or share dedup work across entities in the same file.

## Suggested fix

- Intern per-file reference tokens and store small integer token IDs in line entries.
- Store kind flags (`bare`, `call`, `import`, `dot_receiver`, `dot_member`) alongside token IDs instead of separate owned collections.
- Use empty-line sentinels or sparse line maps so lines without references do not allocate independent `Vec`/`HashSet` values.
- Consider prefix/range indexes for common entity-range queries.
- Build indexes only for languages/files that can reach graph fallback after scope resolution.
- Add a benchmark fixture with large generated files and many nested entities.

## Related issues / PRs

- #248 tracks fallback correctness when scope resolution has already declined a reference.
- #308 introduces the bounded fallback index and should land before this follow-up is attempted.

### LINKED ISSUE #315: Reuse parsed import metadata across TS/JS graph resolution
**Severity:** High
**Component:** parser / graph / TS/JS import resolution
**Affected commands:** `sem impact`, `sem graph`, `sem context`, `sem verify`, MCP graph/impact tools
**Reproduced on:** source inspection plus generated TS fixtures

## Summary

The TS/JS graph path still has two import-resolution pipelines that do overlapping work during cold graph builds and partial rebuilds:

1. `EntityGraph::build` / `build_incremental` builds a graph-level import table.
2. `scope_resolve::resolve_with_scopes_full` walks each AST and extracts imports again for scope-aware resolution.

Recent PR work reduces local hotspots in default and namespace import handling, but it does not remove the duplicate import metadata pass. Large TS/JS repos with many import statements still pay for import discovery twice, and the two pipelines must keep matching semantics for default, named, namespace, barrel, direct-file, and index-file imports.

## Reproduction

A public synthetic fixture is enough to exercise the shape:

```sh
tmp=$(mktemp -d)
cd "$tmp"
git init -q
for i in $(seq 1 2000); do
  printf 'export function f%s() { return %s }\n' "$i" "$i" > "m$i.ts"
  printf 'import { f%s } from "./m%s";\nexport function g%s() { return f%s(); }\n' "$i" "$i" "$i" "$i" > "u$i.ts"
done

/usr/bin/time -p sem graph --no-cache --file-exts .ts --json >/dev/null
```

The current command does not expose phase timings, so the easiest way to verify the duplicated work today is source inspection:

```sh
rg -n "build_import_table|extract_imports_from_ast|resolve_with_scopes_full" crates/sem-core/src/parser
```

## Observed

The graph build creates a graph-level import table and then invokes scope resolution, which extracts import metadata from the same files again. Incremental builds also rebuild the import table from all file paths, even when only a small set of files is stale.

This creates avoidable cold-path work and keeps two import semantics implementations coupled by convention rather than by shared data.

## Expected

A TS/JS file's import metadata should be extracted once per parsed file and reused by both graph-level fallback resolution and scope-aware resolution.

## Root cause

Import metadata is not represented as a first-class per-file product of parsing/extraction. The graph layer asks for one import table, while scope resolution computes its own local import table from ASTs.

## Suggested fix

Introduce a shared import metadata/index layer, for example:

- Build a per-file `ImportIndex` while the file content/tree is already in memory.
- Store enough information for default, named, namespace, barrel, direct file, and index file resolution.
- Pass the index through `PreBuiltLookups` or a sibling input to `resolve_with_scopes_full`.
- Make legacy graph fallback consume the same index for scope-resolved languages.
- Add parity tests proving default/named/namespace resolution picks the same single module candidate in both paths.

## Related issues / PRs

- #111 tracks older cold-build serial bottlenecks and mentions `build_import_table` cost.
- #264 tracks TS import-resolution correctness gaps.
- #309 reduces several TS/JS import-resolution hotspots but intentionally stops short of unifying the import metadata pipeline.

### LINKED ISSUE #316: Add public graph-build phase timings and large generated fixtures
**Severity:** Medium
**Component:** benchmarks / observability / performance regression testing
**Affected commands:** `sem impact`, `sem graph`, `sem context`, `sem verify`, MCP graph/impact tools
**Reproduced on:** repeated performance investigations that required ad hoc timing

## Summary

Recent performance work has found several real graph-build bottlenecks, but the project still lacks a public, repeatable way to measure large-repo behavior by phase. That makes performance regressions hard to catch without relying on private repositories or one-off local scripts.

We need two things:

1. phase timings for graph/cache commands,
2. generated public fixtures large enough to exercise TS/JS imports, scope resolution, fallback indexing, cache load, incremental rebuild, and cache save.

## Reproduction

Today there is no first-class command like this:

```sh
SEM_TIMINGS=1 sem impact f1000 --file f1000.ts --file-exts .ts --json >/dev/null
```

Expected timing buckets would include at least:

- file discovery
- cache open/load
- full entity extraction
- entity lookup/index construction
- import metadata/table construction
- scope resolution
- fallback reference indexing
- fallback edge resolution
- cache save/incremental save
- CLI output serialization

Without these buckets, each investigation has to add temporary instrumentation or infer bottlenecks from source inspection and wall-clock totals.

## Observed

The code has useful tests and some benchmark-like tests, but no public fixture generator or stable phase-timing output for large graph builds. Performance reports therefore tend to mix code reading, local timing, and private-repo observations, which are hard for maintainers and contributors to reproduce.

## Expected

A contributor should be able to run a public benchmark and produce a timing report that is safe to paste into an issue or PR.

## Suggested fix

- Add a generated fixture tool, for example `scripts/gen-large-fixture`, with knobs for:
  - file count,
  - entities per file,
  - import fan-out,
  - namespace/default/named import mix,
  - nested class/function depth,
  - generated large entity bodies.
- Add an ignored integration test or benchmark target that builds the fixture and records wall time for cold cache, warm cache, and one-file incremental rebuild.
- Add machine-readable timings behind an opt-in flag or environment variable, for example:
  - `sem graph --timings --json`
  - `SEM_TIMINGS=json sem impact ...`
- Keep timing output on stderr unless explicitly requested in JSON, so existing stdout contracts remain stable.
- Document the benchmark command in the performance issue template or contributing docs.

## Related issues / PRs

- #105 and #111 were hard to compare over time because phase timings were not built into the command.
- #307, #309, #310, and #312 include targeted runtime improvements; this issue is about preventing the same class of regression from returning unnoticed.

### LINKED ISSUE #317: Add a topology-only cache load path for impact deps and dependents
**Severity:** Medium
**Component:** CLI impact / cache / MCP graph tools
**Affected commands:** `sem impact --deps`, `sem impact --dependents`, MCP impact calls in dependency/dependent modes
**Reproduced on:** source inspection plus warm-cache large-entity fixtures

## Summary

`sem impact --deps` and `sem impact --dependents` only need graph topology: entity IDs/names/types/files/line ranges and edges. Today the shared graph loader always returns `(EntityGraph, Vec<SemanticEntity>)`, and the cache load path selects full entity content, content hashes, structural hashes, parent IDs, and metadata for every entity.

That means topology-only impact queries hydrate all source snippets from SQLite even when they will not inspect entity content.

## Reproduction

Create a warm cache with many large entity bodies:

```sh
tmp=$(mktemp -d)
cd "$tmp"
git init -q
for i in $(seq 1 2000); do
  {
    printf 'export function f%s() {\n' "$i"
    yes '  const payload = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";' | head -200
    printf '  return %s;\n}\n' "$i"
  } > "f$i.ts"
done

git add . && git commit -qm init
sem graph --file-exts .ts >/dev/null
sqlite3 .sem/cache.db 'select sum(length(content)) from entities;'
/usr/bin/time -p sem impact f1000 --file f1000.ts --file-exts .ts --deps --json >/dev/null
/usr/bin/time -p sem impact f1000 --file f1000.ts --file-exts .ts --dependents --json >/dev/null
```

Source inspection:

```sh
rg -n "get_or_build_graph|SELECT id, name, entity_type, file_path, start_line, end_line, content|print_deps|print_dependents" crates/sem-cli/src
```

## Observed

`impact_command` always calls `get_or_build_graph`, which returns full `SemanticEntity` values. `DiskCache::load` selects the `content` column for every entity. `print_deps` and `print_dependents` only use `EntityGraph`.

## Expected

Dependency/dependent impact modes should use a cache path that only loads graph topology. Full entity content should be loaded lazily or only for modes that need it, such as test filtering or context generation.

## Root cause

The cache API does not distinguish graph topology from full semantic-entity payloads, and the impact command does not choose a load mode based on `ImpactMode`.

## Suggested fix

- Add `DiskCache::load_graph_topology` that selects only `id`, `name`, `entity_type`, `file_path`, `start_line`, `end_line`, `parent_id`, and edges.
- Add a graph command helper that can return topology-only data for `ImpactMode::Deps` and `ImpactMode::Dependents`.
- Consider caching a derived `is_test` bit or test-entity index so `ImpactMode::Tests` does not need every content string either.
- Mirror the topology-only path in MCP impact handling.
- Add a regression benchmark where entity content is intentionally large and `--deps` wall time is compared before/after.

## Related issues / PRs

- #125 covered cache indexes/schema versioning.
- #307 improves incremental graph selection; this issue targets warm-cache hydration cost for topology-only queries.

### LINKED ISSUE #318: Make incremental cache updates granular for import metadata and edges
**Severity:** High
**Affected commands:** `sem impact`, `sem graph`, `sem context`, `sem verify`, MCP graph/impact tools
**Reproduced on:** source inspection plus generated large-repo fixtures

## Summary

The incremental cache path still does too much whole-graph work after a small edit. Recent PR work reduces several in-memory scans when choosing files/entities to revisit, but the cache layer still has coarse operations that are expensive on large graphs:

- `load_partial` loads all cached entities and all cached edges, then splits stale vs clean in memory.
- `build_incremental` rebuilds import metadata from all file paths.
- `save_incremental` deletes every edge and reinserts the full graph edge set.

For warm-cache workflows, a one-file edit should be close to proportional to the changed file plus directly affected import/reverse-edge neighborhood, not to total repo edge count.

## Reproduction

A synthetic public fixture can show the shape without relying on private repositories:

```sh
tmp=$(mktemp -d)
cd "$tmp"
git init -q
for i in $(seq 1 3000); do
  printf 'export function f%s() { return %s }\n' "$i" "$i" > "m$i.ts"
  printf 'import { f%s } from "./m%s";\nexport function g%s() { return f%s(); }\n' "$i" "$i" "$i" "$i" > "u$i.ts"
done

git add . && git commit -qm init
sem graph --file-exts .ts >/dev/null
printf '\nexport function extra() { return 1 }\n' >> m1500.ts
/usr/bin/time -p sem impact f1500 --file m1500.ts --file-exts .ts --json >/dev/null
sqlite3 .sem/cache.db 'select count(*) from entities; select count(*) from edges;'
```

Source inspection confirms the coarse operations:

```sh
rg -n "load_partial|build_import_table|DELETE FROM edges|INSERT INTO edges" crates/sem-cli/src/cache.rs crates/sem-core/src/parser/graph.rs
```

## Observed

A partial rebuild still hydrates all cached edge rows and rewrites all edge rows after the graph update. It also rebuilds import data from the full file set instead of persisting and invalidating import metadata per file.

## Expected

A warm-cache edit should:

- read stale-file entities plus the minimal clean context needed for resolution/traversal,
- refresh import metadata only for stale files and files whose import dependencies require recomputation,
- delete/reinsert only affected edge rows,
- keep unchanged clean-clean edges untouched in SQLite.

## Root cause

The cache stores final graph data, but it does not persist per-file import metadata or maintain an edge-update plan. Because affected clean entities can be re-resolved, the current save path takes the safe coarse route: delete all edges and insert the complete graph again.

## Suggested fix

1. Persist per-file import metadata keyed by content hash or mtime, with schema versioning.
2. Add explicit edge indexes if they are not already present in the active schema: `edges(from_entity)` and `edges(to_entity)`.
3. During incremental rebuild, compute `affected_source_ids` and `deleted_ids`.
4. Replace full edge rewrite with targeted deletion, for example:
   - `DELETE FROM edges WHERE from_entity IN affected_source_ids`
   - `DELETE FROM edges WHERE to_entity IN deleted_ids`
   - insert recalculated edges for affected sources
5. Add a large generated fixture benchmark that edits one leaf file and records cache load, graph rebuild, and cache save timings separately.

## Related issues / PRs

- #116 covered stale import-scope correctness in the cache.
- #249 covered a clean-caller edge correctness issue.
- #307 reduces several incremental in-memory scans; this issue tracks remaining cache granularity work after that PR lands.

Iron-Ham:
## Summary

- Optimizes `sem diff --patch` blob/content resolution and formatter JSON/string output to avoid per-file subprocess and allocation hot paths.
- Adds topology-only graph cache loading for `impact --deps` / `--dependents`, plus opt-in CLI phase timings via `SEM_TIMINGS`.
- Makes incremental cache edge updates granular, with safeguards for JS/TS import/re-export invalidation and MCP cache compatibility.
- Adds a public large JS/TS fixture generator and benchmark runner.

Closes #313
Closes #314
Closes #315
Closes #316
Closes #317
Closes #318

## Test plan

- `cargo test --manifest-path crates/Cargo.toml --workspace`
- `node --test scripts/large-js-fixture.test.mjs benchmarks/large-js-fixture/run.test.mjs`
- `cargo build --manifest-path crates/Cargo.toml --release --bin sem`
- `SEM_TIMINGS=json sem graph . --json --file-exts .ts .js` smoke on generated fixture

## Benchmarks

Release binary compared against `/tmp/sem-perf-baseline/sem-before`.

- Generated graph fixture, 1,500 mixed JS/TS files:
  - cold cache: 2822.848 ms before, 3513.053 ms after
  - warm cache: 172.095 ms before, 174.801 ms after
  - one-file incremental: 433.205 ms before, 379.957 ms after
- Generated cached `impact --deps`, 2,500 mixed JS/TS files:
  - average: 220 ms before, 170 ms after
- Generated 1,000-file TS patch through `sem diff --patch --json`:
  - 61220 ms before, 410 ms after
- Detached `/tmp` worktree of a 4,415-file TS/JS repo:
  - cold graph: 5930 ms before, 5830 ms after

## Large-repo graph resolution scaling (follow-on)

A follow-on commit (`2b58693`) targets the cold entity-graph build that this PR disclosed as a regression — the path used by `sem graph` and by the first `sem impact` / `sem context` before the cache is warm. That build scaled super-linearly, making a fresh checkout's first `sem impact` run for many minutes on very large monorepos. Profiling a large real-world TypeScript/JavaScript monorepo (~71,000 files, ~2.7M entities) showed the scope resolver dominating. These changes are performance-only and output-preserving.

- **`resolve_ref` candidate-scan gating.** The global symbol-table scan for same-name entities only ever fed Swift overload disambiguation, so it is now gated behind the presence of Swift call signatures; other languages take a same-file match from a per-file name index in O(1).
- **Pass-2 import table (quadratic → linear).** Each file rescanned the entire prebuilt import table for its own entries — O(files × imports). It is grouped by file once up front, so each file does work proportional to its own imports. This is the largest contributor at high file counts.
- **Unused default-export table.** Per-symbol imports are already resolved on the graph-build path and the JS/TS default-export table is never read there, so building it is now skipped on that path.
- **Parallel default-export builders.** Both default-export table builders were single-threaded loops over every file; they now run as parallel map-reduces with identical output.
- **Per-reference lookup hoisting.** The owning-entity span/children, descendant-range, and consumed-words lookups were re-hashed on every reference; they are now fetched once per entity, removing the HashMap hashing that dominated the hottest loop.
- **Name-keyed per-file import table.** `resolve_ref` allocated a `(path, name)` tuple key per reference; since the per-file table only holds the current file's imports, it is now keyed by name alone.

Headline results, measured primarily in **instructions retired** (`/usr/bin/time -l`, resilient to CPU throttling and machine contention; wall-time is noisier):

| Workload | Instructions retired | Wall |
|---|---|---|
| 32,000-file / ~245k-entity TS/JS subset | **−9.8%** | ~1.3× |
| Second large TS/JS repo (~4,400 files, ~244k entities) | −4.3% | ~1.1× |
| ~71,000-file / ~2.7M-entity monorepo, cold `sem impact` in a fresh worktree | completes full graph build in **~452 s** (then **~6.5 s** warm/cached) | — |

The win grows with file count because it is dominated by the quadratic pass-2 fix. The ~71k-file baseline was previously impractical to run to completion — the removed O(files × imports) rescan is tens of billions of operations in a hot path at that scale — and the optimized build now finishes and warms a cache that serves subsequent calls in seconds.

Output is **byte-identical** before/after: `sem graph --json` produces identical entities and edges (sorted-output diffing) on five real repositories spanning TypeScript, JavaScript, and Rust, and `cargo test --workspace` passes (25 suites). A small **pre-existing** run-to-run nondeterminism (~5 edges out of 533k on a 32k-file repo) in ambiguous method-call resolution via HashMap ordering affects the baseline and this branch identically and is not introduced here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -216,3 +216,7 @@ This prints the full AST with node types.
 ## Questions?
 
 Open an issue or check the existing language configs in `languages.rs` for reference. The simplest configs (Python, Bash) are good starting points. The Elixir config shows the call-based approach.
+
+## Performance Fixtures
+
+For graph-build performance work, use the generated JS/TS fixture benchmark in `benchmarks/large-js-fixture/`. It can regenerate large public fixtures and measure cold cache, warm cache, and one-file incremental rebuild timings without relying on private repositories.
--- a/benchmarks/large-js-fixture/.gitignore
+++ b/benchmarks/large-js-fixture/.gitignore
@@ -1,0 +1,2 @@
+.generated/
+results*.json
--- a/benchmarks/large-js-fixture/README.md
+++ b/benchmarks/large-js-fixture/README.md
@@ -1,0 +1,66 @@
+# Large JS/TS Fixture Benchmark
+
+This benchmark provides a public, repeatable TypeScript/JavaScript fixture for graph-build performance work. It can be paired with `SEM_TIMINGS=1` or `SEM_TIMINGS=json` to collect CLI phase timings while it measures cold cache, warm cache, and one-file incremental runs.
+
+## Generate a Fixture
+
+```bash
+node scripts/large-js-fixture.mjs \
+  --out /tmp/sem-large-js-fixture \
+  --files 1000 \
+  --entities-per-file 12 \
+  --fanout 5 \
+  --import-style-mix named:3,default:1,namespace:1,type:1 \
+  --nested-depth 3 \
+  --body-lines 40 \
+  --language mixed
+```
+
+The generator writes a standalone fixture root with `src/`, `package.json`, `tsconfig.json`, `fixture-manifest.json`, and a marker file. Existing marked fixture roots are replaced for repeatability. Non-empty unmarked directories are refused unless `--force` is supplied.
+
+## Time sem
+
+```bash
+node benchmarks/large-js-fixture/run.mjs \
+  --sem crates/target/debug/sem \
+  --files 1000 \
+  --entities-per-file 12 \
+  --fanout 5 \
+  --import-style-mix named:3,default:1,namespace:1,type:1 \
+  --nested-depth 3 \
+  --body-lines 40 \
+  --language mixed
+```
+
+The runner:
+
+1. regenerates the fixture,
+2. clears the benchmark cache root,
+3. runs a cold-cache command,
+4. runs the same command with a warm cache,
+5. mutates one marker inside the first source file,
+6. runs the same command again for one-file incremental rebuild timing.
+
+By default it runs:
+
+```bash
+sem graph . --json --file-exts .ts .js
+```
+
+It writes machine-readable results to `benchmarks/large-js-fixture/.generated/results.json`. Use `--json-out <path>` to choose another path or `--no-json-out` to print only the terminal summary.
+
+## Useful Knobs
+
+- `--files`: source file count.
+- `--entities-per-file`: top-level generated value entities in each source file.
+- `--fanout`: number of neighboring files imported by each file.
+- `--import-style-mix`: comma-separated style cycle with optional weights. Supported styles are `named`, `default`, `namespace`, `type`, and `side-effect`.
+- `--nested-depth`: nested local function depth inside each top-level entity.
+- `--body-lines`: extra executable lines in each top-level entity body.
+- `--language`: `ts`, `js`, or `mixed`.
+- `--command`: `graph`, `impact`, `context`, or `verify`.
+- `--impact-mode`: `deps`, `dependents`, `tests`, or `all`. The default is `deps` so `impact` benchmarks exercise the topology-only path.
+- `--timeout-ms`: per-run guard for commands under investigation.
+- `--cache-dir`: cache root. Keep it outside `--out` so sem accepts the override.
+
+The default output lives under `benchmarks/large-js-fixture/.generated/`. The generated fixture is initialized as its own Git repository by default so sem treats that directory as the repository root even when the fixture is inside this checkout.
--- a/benchmarks/large-js-fixture/run.mjs
+++ b/benchmarks/large-js-fixture/run.mjs
@@ -1,0 +1,450 @@
+#!/usr/bin/env node
+
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import { spawn } from 'node:child_process';
+import { performance } from 'node:perf_hooks';
+import { fileURLToPath } from 'node:url';
+import {
+  DEFAULT_LARGE_JS_FIXTURE_OPTIONS,
+  generateLargeJsFixture,
+  parseImportStyleMix,
+} from '../../scripts/large-js-fixture.mjs';
+
+const BENCHMARK_DIR = path.dirname(fileURLToPath(import.meta.url));
+const DEFAULT_OUT_DIR = path.join(BENCHMARK_DIR, '.generated', 'repo');
+const DEFAULT_CACHE_DIR = path.join(BENCHMARK_DIR, '.generated', 'cache');
+const DEFAULT_RESULTS_PATH = path.join(BENCHMARK_DIR, '.generated', 'results.json');
+const COMMANDS = new Set(['graph', 'impact', 'context', 'verify']);
+const IMPACT_MODES = new Set(['deps', 'dependents', 'tests', 'all']);
+
+export const LARGE_JS_FIXTURE_BENCHMARK_HELP = `
+Run cold, warm, and one-file incremental sem timings on a generated JS/TS fixture.
+
+Usage:
+  node benchmarks/large-js-fixture/run.mjs [options]
+
+Fixture knobs:
+  --out <dir>                     Fixture root directory.
+  --files <count>                 Number of source files.
+  --entities-per-file <count>     Top-level value entities per file.
+  --fanout <count>                Imported neighbor files per source file.
+  --import-style-mix <mix>        Comma list or weights: named:3,default,namespace,type.
+  --nested-depth <count>          Nested local function depth inside each entity.
+  --body-lines <count>            Extra executable lines in each entity body.
+  --language <ts|js|mixed>        Source extension mix.
+
+Benchmark knobs:
+  --sem <path>                    sem binary to execute. Defaults to SEM_BIN or sem.
+  --cache-dir <dir>               Cache root outside the fixture root.
+  --command <graph|impact|context|verify>
+  --impact-mode <deps|dependents|tests|all>
+  --target-entity <name>          Entity for impact/context.
+  --target-file <path>            Fixture-relative file for impact/context.
+  --context-budget <tokens>       Token budget for context command.
+  --timeout-ms <millis>           Per-run timeout.
+  --json-out <path>               Write machine-readable results.
+  --no-json-out                   Do not write a results file.
+  --no-git-init                   Do not initialize the fixture as its own Git repo.
+  --force                         Allow deleting a non-empty output directory without marker.
+  -h, --help                      Show this help.
+
+Example:
+  node benchmarks/large-js-fixture/run.mjs --files 1000 --entities-per-file 12 --fanout 5 --body-lines 40
+`.trim();
+
+export function parseLargeJsFixtureBenchmarkArgs(argv, env = process.env) {
+  const options = {
+    sem: env.SEM_BIN || 'sem',
+    outDir: DEFAULT_OUT_DIR,
+    cacheDir: DEFAULT_CACHE_DIR,
+    command: 'graph',
+    impactMode: 'deps',
+    contextBudget: 8000,
+    timeoutMs: 120000,
+    jsonOut: DEFAULT_RESULTS_PATH,
+    fileCount: DEFAULT_LARGE_JS_FIXTURE_OPTIONS.fileCount,
+    entitiesPerFile: DEFAULT_LARGE_JS_FIXTURE_OPTIONS.entitiesPerFile,
+    importFanout: DEFAULT_LARGE_JS_FIXTURE_OPTIONS.importFanout,
+    importStyleMix: DEFAULT_LARGE_JS_FIXTURE_OPTIONS.importStyleMix,
+    nestedDepth: DEFAULT_LARGE_JS_FIXTURE_OPTIONS.nestedDepth,
+    bodyLines: DEFAULT_LARGE_JS_FIXTURE_OPTIONS.bodyLines,
+    language: DEFAULT_LARGE_JS_FIXTURE_OPTIONS.language,
+    gitInit: true,
+    force: false,
+  };
+
+  for (let index = 0; index < argv.length; index += 1) {
+    const arg = argv[index];
+
+    switch (arg) {
+      case '-h':
+      case '--help':
+        return { help: true };
+      case '--sem':
+        options.sem = readFlagValue(argv, ++index, arg);
+        break;
+      case '--out':
+      case '-o':
+        options.outDir = readFlagValue(argv, ++index, arg);
+        break;
+      case '--cache-dir':
+        options.cacheDir = readFlagValue(argv, ++index, arg);
+        break;
+      case '--command':
+        options.command = readFlagValue(argv, ++index, arg);
+        break;
+      case '--impact-mode':
+        options.impactMode = readFlagValue(argv, ++index, arg);
+        break;
+      case '--target-entity':
+        options.targetEntity = readFlagValue(argv, ++index, arg);
+        break;
+      case '--target-file':
+        options.targetFile = readFlagValue(argv, ++index, arg);
+        break;
+      case '--context-budget':
+        options.contextBudget = parsePositiveInteger(readFlagValue(argv, ++index, arg), arg);
+        break;
+      case '--timeout-ms':
+        options.timeoutMs = parsePositiveInteger(readFlagValue(argv, ++index, arg), arg);
+        break;
+      case '--json-out':
+        options.jsonOut = readFlagValue(argv, ++index, arg);
+        break;
+      case '--no-json-out':
+        options.jsonOut = null;
+        break;
+      case '--files':
+      case '--file-count':
+        options.fileCount = parsePositiveInteger(readFlagValue(argv, ++index, arg), arg);
+        break;
+      case '--entities-per-file':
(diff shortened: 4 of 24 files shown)


Iron-Ham:
@rs545837 it's already _better_ in many regards (vs `main`) but my aim is to get the performance good enough so we have reasonable cold execution times on a 100k-file repo.

Issue #324, opened 2026-06-05
Low-effort global wins for the graph build: custom allocator and a faster hasher


Iron-Ham:
# Low-effort global wins for the graph build: custom allocator and a faster hasher

## Summary

Two small, low-risk experiments likely recover a meaningful slice of cold-graph-build cost before any of the larger refactors, and are worth measuring first. Neither changes graph algorithms, but both should be scoped carefully because they affect process-wide or map-order behavior.

1. **A non-default global allocator** (e.g. `mimalloc` or `jemalloc`) in the final binary crates.
2. **A non-cryptographic hasher** (e.g. `rustc-hash`/`FxHashMap` or `ahash` with explicit fixed seeding where determinism matters) for the hot graph/resolver maps.

These attack the two costs the profiling pass found most prominent — allocation churn and `HashMap` hashing — without touching resolution logic. They also compose with the bigger issues (interning, per-reference resolution) rather than competing with them.

## 1. Custom global allocator

There is currently no `#[global_allocator]` set, so the build uses the system allocator. The cold build is highly allocation-intensive and many-threaded (rayon across pass-1/pass-2, millions of short-lived `String`/`Vec` allocations), which is exactly the workload where the default system allocator — particularly on macOS — can underperform allocators like `mimalloc` or `jemalloc`.

Apply this at the binary boundary (`sem-cli` and, if measured useful, `sem-mcp`) rather than in `sem-core`: `sem-core` is a library, and a library-level global allocator would take allocator choice away from downstream embedders. A binary-scoped experiment is still a small change:

```rust
#[global_allocator]
static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
```

This commonly buys high-single-digit-to-low-double-digit percent wall-time on allocation-heavy Rust workloads and can reduce RSS fragmentation, which independently helps the memory pressure described in #322 (possibly reducing the urgency of that refactor). It is purely an experiment to measure; revert or gate per target if it doesn't pay off on the platforms we ship.

## 2. Faster hasher for hot maps

The resolver uses `std::collections::HashMap` everywhere. `hash_one` was the single hottest leaf in the pass-2 profile, and the std default is SipHash 1-3 today (DoS-resistant, but slow for many short string keys). The hot maps in `resolve_ref` and graph assembly key on short reference names and entity IDs; a non-cryptographic hasher (`FxHashMap` from `rustc-hash`, or carefully configured `ahash`) is typically several times faster for these keys.

[...]

Issue #322, opened 2026-06-05
Cold graph build holds the repo's source roughly twice over, plus all parse trees


Iron-Ham:
# Cold graph build holds the repo's source roughly twice over, plus all parse trees

## Summary

The cold entity-graph build keeps the repository's source text resident roughly twice over, plus every file's tree-sitter tree, at its peak. This is likely the largest *byte* contributor to peak RSS (~27 GB on a large monorepo of ~71k files / ~2.7M entities). Each `SemanticEntity` owns a `content: String` copy of its source span, and the builder additionally retains every file's full text and parsed tree during scope resolution. Reducing this lowers peak memory and the allocator pressure that contributes to build time.

## Background / evidence

`SemanticEntity` (`crates/sem-core/src/model/entity.rs`) carries `content: String` (the entity's source) plus `content_hash: String` and an optional `structural_hash: String` — i.e. up to three per-entity strings, the largest being `content`. The text is also persisted to the cache (`content TEXT NOT NULL`).

Two whole-repo-scale allocations coexist at the peak:

- **Entity `content`.** Entities tile their files, so the sum of all `content` is *on the order of* one full copy of the repo's source. It is not exactly 1×: nested entities overlap (a class's `content` contains its methods' `content`), which can push it above 1× in deeply-nested files, while top-level trivia not owned by any entity can pull it below. Treat "~1×" as an estimate, not a bound. This copy is resident for the whole build.
- **Parse trees + file text.** `EntityGraph::build` (`parser/graph.rs`) collects `parsed_files: Vec<(String, String, tree_sitter::Tree)>` — every file's path, full text, and tree — and moves the whole vector into the scope resolver, which holds it across pass 1 / pass 1b / pass 2. Tree-sitter trees are large relative to source, so this is a second whole-repo-scale allocation. It is freed when scope resolution returns (before the `build()` bag-of-words phase), so the tree peak coincides with the scope-resolution phase; the entity-`content` peak persists throughout.

Net: at the scope-resolution peak, source is resident ~2× (entity `content` + `parsed_files` text) plus all trees. There is typically enough headroom on a large-memory host (no swapping), so this is not a hard failure today — but it scales linearly with repo size and bites smaller-memory and CI environments, and the allocator churn has a wall-time cost even with headroom.

## Proposal (candidate directions)

[...]

Issue #321, opened 2026-06-05
Scope-resolver pass 2 dominates cold build: reduce per-reference work


Iron-Ham:
# Scope-resolver pass 2 dominates cold build: reduce per-reference work

## Summary

In the cold entity-graph build, the scope resolver's pass 2 (`resolve_with_scopes_full` in `crates/sem-core/src/parser/scope_resolve.rs`) is the dominant cost — roughly 60% of the cold build's *instructions retired* on a large monorepo (measured at ~32k–71k files; the share moves with repo size and the parse/IO mix). It walks every reference in every entity and runs a chain of `HashMap` lookups (`resolve_ref` and its helpers) to resolve each one. With ~2.7M entities and tens of millions of references, even a few hundred nanoseconds per reference adds up to minutes.

There are two independent levers on this loop: reduce the **number** of lookups per reference (this issue), and reduce the **cost** of each lookup (#324 — the loop hashes short reference names with the default randomly-seeded SipHash, which is the profile's hottest leaf). Identifier interning does *not* help here, because these lookups key on names, not entity IDs.

## Background / evidence

For each entity, pass 2 iterates the references in its line range and calls `resolve_ref`, which attempts resolution in order: local-binding check, scope-chain walk, per-file import table, then global symbol-table fallback — plus method-call paths consulting `class_members`, `owner_members`, `instance_attr_types`, and enclosing-class/type caches. That is tens of `HashMap` lookups across `resolve_ref` and its `*_cached` helpers, the exact count depending on which `AstRefKind` branch runs.

Profiling the cold build showed `core::hash::BuildHasher::hash_one` as the hottest leaf, concentrated in this loop. Prior work in #319 already removed the worst structural offenders (a global same-name candidate scan, an O(files × imports) per-file import-table rescan, a per-reference key-string allocation in the resolver import lookup, and per-reference re-hashing of owning-entity/descendant lookups). What remains is the inherent volume: `resolve_ref` runs roughly once per reference, and each invocation still performs several map lookups.

## Proposal (candidate directions, not prescriptive)

[...]

Issue #320, opened 2026-06-05
Intern entity identifiers to cut graph-build memory and allocation churn


Iron-Ham:
# Intern entity identifiers to cut graph-build memory and allocation churn

## Summary

Entity IDs are long, fully-qualified strings (e.g. `app/services/user.ts::function::createUser`) stored as owned `String`s in many structures at once. On a large monorepo (~71k files, ~2.7M entities) this duplication is a major contributor to the cold graph build's peak memory (which reaches ~27 GB) and to allocation churn from cloning IDs into multiple maps. Replacing the runtime identifier with an interned integer handle (the string materialized only at output and cache boundaries) should cut peak RSS and reduce the clone/alloc traffic during graph assembly.

This is primarily a **memory and allocation** optimization. It is *not* expected to speed up the per-reference resolution loop — that loop hashes short reference *names*, not entity IDs (see #321 and #324 for the resolution-speed levers).

## Background / evidence

In `crates/sem-core/src/parser/graph.rs`, each entity ID is duplicated across several long-lived structures:

- `EntityGraph.entities: HashMap<String, EntityInfo>` — the ID is both the map key and `EntityInfo.id`, plus the optional `EntityInfo.parent_id`.
- `EntityGraph.dependents` and `EntityGraph.dependencies: HashMap<String, Vec<String>>` — every edge contributes cloned IDs to the forward/reverse indexes.
- `EntityGraph.edges: Vec<EntityRef>` where `EntityRef { from_entity: String, to_entity: String, .. }`.
- `symbol_table: HashMap<String /*name*/, Vec<String /*entity IDs*/>>` — the IDs are the values (the key is the entity *name*).

So an entity with incident edge degree *d* is stored in several fixed places, plus O(*d*) additional edge/index copies. For example, an outgoing edge stores the source ID in `EntityRef::from_entity` and in the target's `dependents` vector; an incoming edge stores it in `EntityRef::to_entity` and in the source's `dependencies` vector. The exact count varies with in/out degree, parent status, and map-key reuse, but at ~2.7M entities and ~1.7M edges this is still millions of duplicated identifier strings. It also drives allocation churn: graph assembly clones these IDs repeatedly — `EntityGraph::from_parts` and the edge-index build clone `from_entity`/`to_entity` into the `dependents`/`dependencies` maps, and edge de-duplication keys a `HashSet` on cloned `(from, to)` ID pairs.

[...]

PR #326, opened 2026-06-06
Reduce graph memory and cache overhead

Iron-Ham:
## Summary

- Streams `sem graph --json` output and adds a fresh-cache SQLite streaming path so graph JSON no longer duplicates the full graph in memory.
- Avoids repo-wide fallback parsing during graph scans, prunes default-excluded directories before descent, and removes the extra NUL-byte probe for known source extensions.
- Bounds huge full-build memory by avoiding retained parse trees above 20k files, resolving scope in chunks, and building fallback reference indexes per file.
- Cuts scope-resolution CPU/allocation churn by precomputing deterministic return-type lookups once, avoiding per-node AST child vector allocations, and skipping Swift signature scans for non-Swift chunks.
- Reduces incremental-cache peak memory by streaming cached entity rows directly into clean/stale buckets.
- Stores source-file content fingerprints in the cache so mtime-only churn can refresh mtimes without rebuilding the graph.
- Adds `mimalloc` for the CLI and MCP binaries.

## Stacking note

This branch is intentionally based on #325 (`Iron-Ham/fix-deterministic-graph-output`) because this work builds on the deterministic graph output and earlier performance work in the open stack. The PR is still pointed at upstream `main` so reviewers can see the eventual merge target, but GitHub cannot hide the stacked #319/#325 commits and diff while targeting `main`. The new implementation in this PR is the single top commit: `ffa9398 perf: reduce graph memory and cache overhead`.

## Performance issues considered

This implements the best immediate wins from #322, #314, #324, #316, #317, and #321: reduce retained graph/source/parse-tree memory, compact fallback reference work, add low-effort allocator and scope-resolution wins, stream graph JSON from cache, and keep public benchmark coverage. #318 is partially addressed by targeted edge rewrites, cache streaming, and content-aware freshness checks, but remains open for full per-file import metadata and cache granularity. #320 and #315 remain deeper follow-ups for entity interning and persisted TS/JS import metadata; this PR only takes the parts that had a direct, measured win in the current graph path.

## Benchmarks

Release binary compared against a clean PR #325 baseline where noted.

- Generated 5,000-file JS/TS graph fixture, `sem graph . --json --file-exts .ts .js`:
  - cold cache: 10422.991 ms before, 7804.303 ms after
  - warm cache: 944.993 ms before, 441.568 ms after
  - one-file incremental: 2275.921 ms before, 1391.307 ms after
- Generated 5,000-file JS/TS fixture, cold no-cache graph RSS (`/usr/bin/time -l`):
  - max resident set size: ~10.02 GB before, ~7.88 GB after
  - after wall time: 6.66 s
- Large 100k+ file JS/TS monorepo, `sem graph . --file-exts .ts .tsx .js .jsx .mjs .cjs` with isolated cache:
  - before: earlier probe hit ~12.5 GB RSS and was stopped
  - cold cache after: 59.63 s wall, 6.05 GB max RSS, 660741 entities, 1647490 edges
  - warm cache after: 3.05 s wall, 1.54 GB max RSS, same graph counts
  - one-file content incremental after: 9.96 s wall, 3.91 GB max RSS, same graph counts
  - 500-file mtime-only churn after: 3.17 s wall, 1.54 GB max RSS, same graph counts, stayed on `cache_topology_load`

The largest confirmed win is peak memory/RSS. Warm cache and one-file incremental are also substantially faster because `graph --json` can avoid entity-body loads and giant JSON allocations, the full graph path no longer repeats the deterministic return-type index for every file, and mtime-only churn no longer forces expensive incremental rebuilds.

## Graph correctness validation

Compared `sem graph --no-cache --json --file-exts .ts .tsx .js .jsx .mjs .cjs` against the PR #325 binary on a 4.4k-file JS/TS repo with a similar monorepo shape. The canonical graph output matched exactly after normalizing to graph-relevant fields:

- entities: 90591 before, 90591 after
- edges: 141938 before, 141938 after
- canonical entity tuples: no mismatches
- canonical edge tuples: no mismatches

The raw JSON SHA changed only because object key order differs; the graph structure and ordered canonical entity/edge tuples match.

## Test plan

- `cargo test --manifest-path crates/Cargo.toml --workspace`
- `cargo build --manifest-path crates/Cargo.toml --release -p sem-cli -p sem-mcp`
- `node benchmarks/large-js-fixture/run.mjs --sem ./crates/target/release/sem --files 5000 --fanout 3 --timeout-ms 120000 --no-json-out`
- `/usr/bin/time -l ./crates/target/release/sem graph . --no-cache --json --file-exts .ts .js` on the generated 5k fixture
- `sem graph` cold, warm, and one-file incremental runs on a large 100k+ file JS/TS monorepo with an isolated cache
- PR #325 vs this branch canonical graph comparison on a 4.4k-file JS/TS repo
- 500-file mtime-only churn check on a large 100k+ file JS/TS monorepo

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -889,6 +889,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.49"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a45a52f43e1c16f667ccfe4dd8c85b7f7c204fd5e3bf46c5b0db9a5c3c0b8e9"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "libsqlite3-sys"
 version = "0.30.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -975,6 +984,15 @@ name = "memchr"
 version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
+
+[[package]]
+name = "mimalloc"
+version = "0.1.52"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d4139bb28d14ad1facf21d5eb8825051b326e172d216b39f6d31df53cc97862"
+dependencies = [
+ "libmimalloc-sys",
+]
 
 [[package]]
 name = "miniz_oxide"
@@ -1378,6 +1396,7 @@ dependencies = [
  "colored",
  "git2",
  "ignore",
+ "mimalloc",
  "openssl-sys",
  "rusqlite",
  "sem-core",
@@ -1438,6 +1457,7 @@ dependencies = [
  "git2",
  "ignore",
  "lru",
+ "mimalloc",
  "openssl",
  "rmcp",
  "rusqlite",
--- a/crates/sem-cli/Cargo.toml
+++ b/crates/sem-cli/Cargo.toml
@@ -27,6 +27,7 @@ openssl-sys = { version = "0.9", optional = true }
 serde_json = "1"
 rusqlite = { version = "0.32", features = ["bundled"] }
 clap_complete_command = "0.6.1"
+mimalloc = "0.1"
 
 [package.metadata.binstall]
 pkg-url = "{ repo }/releases/download/v{ version }/sem-{ target }{ archive-suffix }"
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, HashSet};
+use std::io::Write;
 use std::path::Path;
 
 use rusqlite::{params, Connection};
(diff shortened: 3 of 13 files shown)


Iron-Ham:
cc: @rs545837 this is a push that takes performance into "plausible" territory for ~100k+ file repos.

rs545837:
@Iron-Ham — https://github.com/Ataraxy-Labs/sem/pull/319 and https://github.com/Ataraxy-Labs/sem/pull/325 are merged. https://github.com/Ataraxy-Labs/sem/pull/326 has conflicts now, needs a rebase.

Iron-Ham:
@rs545837 updated

rs545837:
Hey @Iron-Ham — this stack looks great, merging all four now. Two small follow-up perf items noticed during review:

**1. Batch `refresh_file_fingerprint` updates (this PR)**

`has_fresh_complete_cache` calls `refresh_file_fingerprint` per file inside the loop — each one is a standalone `self.conn.execute(UPDATE...)` without a wrapping transaction. SQLite autocommits each UPDATE with a separate fsync. For repos with many mtime-only changes (branch switches, CI cache restores), this could be hundreds of fsyncs during a freshness check.

Fix: collect files needing fingerprint refresh during the loop, then issue all updates in a single `BEGIN...COMMIT` at the end.

**2. Per-reference `(String, String)` allocation in bag-of-words path (pre-existing)**

In `graph.rs` (~line 1070 and ~1880), the bag-of-words resolution path still allocates a `(entity.file_path.clone(), ref_name.to_string())` key tuple for every reference word:

```rust
let import_key = (entity.file_path.clone(), ref_name.to_string());
if let Some(import_target_id) = import_table.get(&import_key) {
```

The scope resolver path already fixed this by re-keying to `HashMap<&str, &str>` (scope_resolve.rs:918-924). The same approach would work here — since the bag-of-words loop processes one entity at a time, a per-entity `HashMap<&str, &str>` view of the import table filtered to the current file would eliminate millions of allocations on large repos.

Neither of these blocks the merge — just noting for follow-up.

Iron-Ham:

Tracked the two follow-ups from https://github.com/Ataraxy-Labs/sem/pull/326#issuecomment-4640277450 as:

- #342
- #343

PR #338, opened 2026-06-08
feat: add D language support

baruch:
## Summary
  - Adds D language support behind a new `lang-d` feature flag, wired into `grammar-all`
  - Uses `tree-sitter-d` 0.8.2 for parsing; registers `.d` and `.di` extensions and the `d` / `dlang` language aliases
  - Extracts modules, functions, classes, structs, interfaces, unions, enums, templates, mixin templates, constructors/destructors/postblits, aliases, unittests, and top-level variables/manifest constants
  - Suppresses nested entity extraction for locals inside functions, constructors, destructors, and `unittest` blocks so only top-level entities surface

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1389,7 +1389,7 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 
 [[package]]
 name = "sem-cli"
-version = "0.7.0"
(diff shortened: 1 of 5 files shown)


rs545837:
Thanks for opening the PR.

rs545837:
Merged manually after rebasing onto main to resolve conflicts from Clojure/Elm PRs. Thanks for the contribution!

PR #344, opened 2026-06-08
Batch cache fingerprint refreshes

### LINKED ISSUE #342: Batch cache fingerprint refresh updates

## Summary

Cache freshness checks refresh mtime/content fingerprints one file at a time when a file is content-identical but has changed mtimes. Each refresh calls `refresh_file_fingerprint`, which executes an `UPDATE` outside an explicit transaction. Repos with many mtime-only changes, such as branch switches or CI cache restores, can turn a freshness check into many independent SQLite writes.

Source follow-up: https://github.com/Ataraxy-Labs/sem/pull/326#issuecomment-4640277450

## Current Code

- CLI freshness loop: https://github.com/Ataraxy-Labs/sem/blob/cc69f3b73b297eeeeb6ee8b251a6c8f4954f8470/crates/sem-cli/src/cache.rs#L496-L512
- CLI per-file update helper: https://github.com/Ataraxy-Labs/sem/blob/cc69f3b73b297eeeeb6ee8b251a6c8f4954f8470/crates/sem-cli/src/cache.rs#L941-L947
- MCP freshness loops: https://github.com/Ataraxy-Labs/sem/blob/cc69f3b73b297eeeeb6ee8b251a6c8f4954f8470/crates/sem-mcp/src/cache.rs#L696-L714 and https://github.com/Ataraxy-Labs/sem/blob/cc69f3b73b297eeeeb6ee8b251a6c8f4954f8470/crates/sem-mcp/src/cache.rs#L830-L843
- MCP per-file update helper: https://github.com/Ataraxy-Labs/sem/blob/cc69f3b73b297eeeeb6ee8b251a6c8f4954f8470/crates/sem-mcp/src/cache.rs#L878-L883

## Candidate Fix

Collect files that need fingerprint refresh during the freshness loop, then apply all updates in a single transaction after freshness has been confirmed. Keep the cache stale/failure behavior explicit if an update fails.

Apply the same pattern to both CLI and MCP cache paths, or consolidate the shared behavior if there is already a suitable common helper boundary.

## Acceptance Criteria

- Mtime-only churn performs one batched transaction per freshness check instead of one standalone update per refreshed file.
- CLI and MCP cache freshness paths use the same batching behavior.
- Existing cache correctness tests pass, with a focused regression test or instrumentation covering multiple `FreshWithUpdatedFingerprint` files.

Iron-Ham:
## Summary
- Batch mtime-only file fingerprint refreshes through a shared SQLite transaction helper instead of writing once per file.
- Apply the batched refresh path to CLI and MCP full/topology/partial cache freshness checks.
- Expand cache tests for multi-file mtime-only refreshes, mixed stale + mtime-only partial loads, and rollback on batched refresh failure.

Fixes #342

## Metrics
Benchmark used a release `sem` binary with 5 isolated runs. Each run generated 5,000 Rust files, primed the cache with `sem graph --file-exts rs`, touched all files without changing content, then timed the cache-hit `sem graph --file-exts rs` path with `SEM_TIMINGS=json`.

| Metric | Baseline median | After median | Change |
| --- | ---: | ---: | ---: |
| `cache_topology_load` phase | 134.16 ms | 82.28 ms | -38.7% |
| total command time | 150.49 ms | 100.77 ms | -33.0% |

Raw `cache_topology_load` samples:
- Baseline: 135.09, 134.16, 131.44, 127.76, 134.95 ms
- After: 82.28, 85.51, 84.35, 77.67, 82.22 ms

## Test Plan
- `cargo test -p sem-cli cache::tests`
- `cargo test -p sem-mcp cache::tests`
- `cargo test --workspace`

### DIFF
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -471,15 +471,14 @@ impl DiskCache {
             return false;
         }
 
-        let mut stmt = self
-            .conn
-            .prepare("SELECT path, mtime_secs, mtime_nanos, content_hash FROM files")
-            .ok();
-        let Some(ref mut stmt) = stmt else {
-            return false;
-        };
-        let cached_mtimes: HashMap<String, (i64, i64, Option<String>)> =
-            match stmt.query_map([], |row| {
+        let cached_mtimes: HashMap<String, (i64, i64, Option<String>)> = {
+            let Ok(mut stmt) = self
+                .conn
+                .prepare("SELECT path, mtime_secs, mtime_nanos, content_hash FROM files")
+            else {
+                return false;
+            };
+            let cached_mtimes = match stmt.query_map([], |row| {
                 Ok((
                     row.get::<_, String>(0)?,
                     (
@@ -492,7 +491,10 @@ impl DiskCache {
                 Ok(rows) => rows.filter_map(|r| r.ok()).collect(),
                 Err(_) => return false,
             };
+            cached_mtimes
+        };
 
+        let mut fingerprint_refreshes = Vec::new();
         for file in files {
             if shared_cache::is_manifest_file_name(file) {
                 continue;
@@ -508,12 +510,18 @@ impl DiskCache {
                     nanos,
                     content_hash,
                 }) => {
-                    self.refresh_file_fingerprint(file, secs, nanos, &content_hash);
+                    fingerprint_refreshes.push(shared_cache::FileFingerprintRefresh {
+                        path: file.clone(),
+                        mtime_secs: secs,
+                        mtime_nanos: nanos,
+                        content_hash,
+                    });
                 }
                 Some(shared_cache::FileFreshness::Stale) | None => return false,
             }
         }
 
+        shared_cache::refresh_file_fingerprints_best_effort(&self.conn, &fingerprint_refreshes);
         true
     }
 
@@ -572,6 +580,7 @@ impl DiskCache {
             .ok()?
             .filter_map(|r| r.ok())
             .collect();
+        drop(stmt);
 
         if cached_files.is_empty() {
             return None;
@@ -587,6 +596,7 @@ impl DiskCache {
         // Find stale source files: mtime differs or not in cache
         let mut stale_source_files: Vec<String> = Vec::new();
         let mut stale_current_file_count = 0;
+        let mut fingerprint_refreshes = Vec::new();
         for file in &source_files {
             match cached_files.get(file.as_str()) {
                 Some((secs, nanos, content_hash)) => {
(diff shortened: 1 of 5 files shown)


Iron-Ham:
Review finding: cache hit now depends on persisting an auxiliary fingerprint refresh.

`FreshWithUpdatedFingerprint` means the file content hash still matches, so the cached graph/topology is valid even if the mtime/content_hash refresh cannot be written. This PR changes the previous best-effort behavior: the new batched refresh is now part of the cache-hit decision (`refresh_file_fingerprints(...).is_ok()` / `.ok()?`). A transient SQLite write, lock, or commit failure can therefore turn an otherwise valid full/topology/partial cache hit into a miss and force a rebuild.

I would preserve the prior semantics: attempt the batched refresh, but do not fail the cache load solely because the auxiliary fingerprint update failed. If strict refresh persistence is intentional, please add a caller-level regression test documenting that cache loads should miss on refresh-write failure.

Verified by review plus cache/workspace test runs; the batching itself is otherwise scoped cleanly and the new rollback test is useful.

Iron-Ham:
Follow-up after the latest update: this resolves my cache-hit concern.

The refresh path is now explicitly best-effort via `refresh_file_fingerprints_best_effort`, so a failed auxiliary fingerprint write no longer invalidates an otherwise valid cache hit. The added `cache_loads_ignore_fingerprint_refresh_failure` regression test covers full load, topology load, and partial load behavior under a forced SQLite update failure.

I verified the updated PR head with:

```text
cargo test -p sem-cli cache_loads_ignore_fingerprint_refresh_failure
cargo test -p sem-mcp cache_loads_ignore_fingerprint_refresh_failure
```

Both passed.

rs545837:
Nice work — I reproduced the benchmark locally and the results are actually better than reported:

| Metric | Baseline | PR | Change |
|---|---:|---:|---:|
| `cache_topology_load` median | 105.9 ms | 48.3 ms | **-54%** |
| total command time median | 119.5 ms | 61.1 ms | **-49%** |

I also isolated where the remaining time goes by comparing a pure cache hit (no files touched) vs the all-touched scenario:

| Scenario | `cache_topology_load` |
|---|---:|
| Pure cache hit (mtimes match) | ~6 ms |
| All 5K files touched | ~48 ms |

So after this PR, ~42ms of the 48ms is sequential filesystem I/O — 5000 calls to `stat()` + `read()` + xxhash in the `has_fresh_cache` / `load_graph_topology` freshness loop. The SQLite read side is only ~6ms now.

One potential follow-up: the freshness checks are embarrassingly parallel since each `file_freshness()` call is independent. A `rayon::par_iter()` over the file list (sem-core already has rayon as an optional dep) could cut that 42ms significantly on multi-core machines.

PR #345, opened 2026-06-08
Avoid import key allocations in graph resolution

### LINKED ISSUE #343: Avoid per-reference import key allocations in bag-of-words resolution

## Summary

The bag-of-words graph resolution path still allocates a `(String, String)` import lookup key for every reference word:

```rust
let import_key = (entity.file_path.clone(), ref_name.to_string());
if let Some(import_target_id) = context.import_table.get(&import_key) {
```

On large repos this can allocate millions of short-lived strings. The scope resolver already avoids the equivalent per-reference import lookup allocation with borrowed keys; this issue tracks applying the same idea to the bag-of-words path.

Source follow-up: https://github.com/Ataraxy-Labs/sem/pull/326#issuecomment-4640277450

Related: #321 notes this as an adjacent bag-of-words cost but leaves it for separate tracking.

## Current Code

- Bag-of-words import lookup allocation: https://github.com/Ataraxy-Labs/sem/blob/cc69f3b73b297eeeeb6ee8b251a6c8f4954f8470/crates/sem-core/src/parser/graph.rs#L928-L955

## Candidate Fix

Avoid constructing an owned tuple for each reference. One possible shape is to build a borrowed per-file view of the import table while processing an entity, such as a `HashMap<&str, &str>` from imported name to target id for `entity.file_path`, then look up `ref_name` directly.

Keep the lookup lifetime local to the entity or file pass so borrowed keys never outlive the source import table.

## Acceptance Criteria

- The bag-of-words path no longer clones `entity.file_path` or allocates `ref_name.to_string()` per reference lookup.
- Graph output is unchanged on representative repos.
- A perf check on a large repo or generated fixture shows reduced allocation churn in the bag-of-words phase.
- Full test suite passes.

Iron-Ham:
## Summary
- Build a borrowed import lookup by file before bag-of-words reference resolution.
- Replace the per-reference `(file_path.clone(), ref_name.to_string())` lookup key with borrowed name lookups.
- Borrow dot-chain consumed words in `resolve_entity_references` instead of allocating owned strings for each consumed `member`/`receiver`.
- Add an ignored allocation benchmark fixture for reproducing the import lookup metric.

Closes #343

## Metrics
Command: `RAYON_NUM_THREADS=1 cargo test --manifest-path crates/Cargo.toml -p sem-core --test bow_import_lookup_bench --release -- --ignored --nocapture`

Fixture: 2,000 TypeScript named imports and 2,000 caller references that resolve through the bag-of-words import fallback. Baseline and branch were run on the same machine/profile.

| Metric | `origin/main` | This branch | Change |
| --- | ---: | ---: | ---: |
| Allocation calls | 473,534 | 461,546 | -11,988 (-2.53%) |
| Allocated bytes | 64,353,075 | 64,484,745 | +131,670 (+0.20%) |
| Mean elapsed | 65.065 ms | 68.126 ms | noisy/no meaningful speed claim |
| Entities | 4,000 | 4,000 | unchanged |
| Edges | 2,000 | 2,000 | unchanged |
| Resolved import edges | 2,000 | 2,000 | unchanged |

Baseline elapsed samples: 64.959 ms, 64.186 ms, 66.049 ms.
Branch elapsed samples: 65.350 ms, 64.240 ms, 74.787 ms.

Relative to the prior PR metric, the borrowed `consumed_words` follow-up saves another 4,000 allocation calls and 88,000 allocated bytes in this fixture. The allocation claim is specifically reduced allocation-call churn in the hot import lookup/reference-resolution path. Total allocated bytes remain slightly higher than `origin/main` in this synthetic one-reference-per-import fixture because the borrowed per-file index allocates bucket storage once.

## Test Plan
- `sem impact --entity-id 'crates/sem-core/src/parser/graph.rs::function::resolve_entity_references' --json`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core import -- --nocapture`
- `cargo test --manifest-path crates/Cargo.toml -p sem-core clojure -- --nocapture`
- `RAYON_NUM_THREADS=1 cargo test --manifest-path crates/Cargo.toml -p sem-core --test bow_import_lookup_bench --release -- --ignored --nocapture`
- `cargo test --manifest-path crates/Cargo.toml`

Note: `cargo test` updates stale local crate versions in `crates/Cargo.lock`; the lockfile was restored so this PR stays scoped. `cargo test --manifest-path crates/Cargo.toml --locked ...` fails for the same pre-existing stale-lockfile reason.

### DIFF
--- a/crates/sem-core/src/parser/graph.rs
+++ b/crates/sem-core/src/parser/graph.rs
@@ -681,10 +681,25 @@ fn direct_reference_line_ranges(
     ranges
 }
 
+type ImportsByFile<'a> = HashMap<&'a str, HashMap<&'a str, &'a str>>;
+
+fn build_imports_by_file<'a>(
+    import_table: &'a HashMap<(String, String), String>,
+) -> ImportsByFile<'a> {
+    let mut imports_by_file: ImportsByFile<'a> = HashMap::new();
+    for ((file_path, import_name), target_id) in import_table {
+        imports_by_file
+            .entry(file_path.as_str())
+            .or_default()
+            .insert(import_name.as_str(), target_id.as_str());
+    }
+    imports_by_file
+}
+
 struct ReferenceResolutionContext<'a> {
     symbol_table: &'a HashMap<String, Vec<String>>,
     entity_map: &'a HashMap<String, EntityInfo>,
-    import_table: &'a HashMap<(String, String), String>,
+    imports_by_file: &'a ImportsByFile<'a>,
     scope_consumed_words: &'a HashMap<String, HashSet<String>>,
     child_ranges_by_parent: &'a HashMap<&'a str, Vec<ChildRange<'a>>>,
     child_line_ranges: &'a HashMap<String, Vec<(usize, usize)>>,
@@ -833,11 +848,6 @@ fn resolve_entity_references(
         direct_reference_line_ranges(entity, fallback_end_line, context.child_line_ranges);
 
     let mut entity_edges = Vec::new();
-    let mut consumed_words = context
-        .scope_consumed_words
-        .get(&entity.id)
-        .cloned()
-        .unwrap_or_default();
 
     let reference_index =
         if entity_requires_content_span_filter(entity, context.child_ranges_by_parent) {
@@ -878,6 +888,11 @@ fn resolve_entity_references(
             })
             .collect(),
     };
+    let mut consumed_words: HashSet<&str> = context
+        .scope_consumed_words
+        .get(&entity.id)
+        .map(|set| set.iter().map(String::as_str).collect())
+        .unwrap_or_default();
 
     for (receiver, member, position) in &dot_chains {
         if consumed_words.contains(*member) {
@@ -906,7 +921,7 @@ fn resolve_entity_references(
                                 target_id.to_string(),
                                 RefType::Calls,
                             ));
-                            consumed_words.insert(member.to_string());
+                            consumed_words.insert(*member);
                             break;
                         }
                     }
@@ -924,15 +939,15 @@ fn resolve_entity_references(
                             target_id.to_string(),
                             RefType::Calls,
                         ));
-                        consumed_words.insert(member.to_string());
-                        consumed_words.insert(receiver.to_string());
+                        consumed_words.insert(*member);
+                        consumed_words.insert(*receiver);
                         break;
                     }
                 }
             }
         }
         if entity_edges.len() == edge_count_before {
-            consumed_words.insert(member.to_string());
+            consumed_words.insert(*member);
         }
     }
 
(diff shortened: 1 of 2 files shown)


Iron-Ham:
Blocking finding: this branch is merge-clean but compile-broken against current `origin/main`.

I reproduced this by merging the PR head into current `origin/main` in a temp worktree and running `cargo check -p sem-core` from `crates/`. The merge succeeds textually, but compilation fails in `sem-core/src/parser/graph.rs`:

```text
error[E0609]: no field `import_table` on type `&ReferenceResolutionContext<'_>`
    --> sem-core/src/parser/graph.rs:1050:53
```

The PR replaces `ReferenceResolutionContext::import_table` with `imports_by_file`, and converts the ordinary import lookup path. Current main also has the Clojure namespace-qualified `alias/name` path, which still calls `context.import_table.get(...)` after the merge. That path needs the same `imports_by_file` lookup treatment.

PR-head tests pass, but this should be fixed before merge because the integrated tree does not compile.

Iron-Ham:
Follow-up after the latest update: the previous post-merge compile blocker is resolved.

I re-ran the same integration check that failed before by merging the updated PR head into current `origin/main` in a temp worktree and running:

```text
cargo check -p sem-core
cargo test -p sem-core clojure -- --nocapture
```

`cargo check -p sem-core` now passes, and the Clojure-focused tests passed as well. The Clojure `alias/name` path now uses the per-file import map instead of the removed `context.import_table` field.

rs545837:
The borrowed import lookup is clean and the benchmark is solid. Verified locally: allocation counts match your numbers exactly (473,534 → 465,546).

One follow-up that stacks on top of this: `consumed_words` in `resolve_entity_references` is currently `HashSet<String>`, which means every `consumed_words.insert(member.to_string())` in the dot-chain loop allocates. Since `member` and `receiver` are already `&str` borrowed from `dot_chains` (which outlives `consumed_words`), the set can be `HashSet<&str>` instead:

```rust
// Before
let mut consumed_words = context
    .scope_consumed_words
    .get(&entity.id)
    .cloned()
    .unwrap_or_default();

// After
let mut consumed_words: HashSet<&str> = context
    .scope_consumed_words
    .get(&entity.id)
    .map(|set| set.iter().map(String::as_str).collect())
    .unwrap_or_default();
```

Then the three insert sites become `.insert(*member)` / `.insert(*receiver)` instead of `.insert(member.to_string())` / `.insert(receiver.to_string())`.

Benchmarked on the same fixture, saves an additional 4,000 allocation calls on top of your change:

| Metric | `main` | This PR | + borrowed consumed_words |
| --- | ---: | ---: | ---: |
| Allocation calls | 473,534 | 465,546 (-1.69%) | 461,546 (-2.53%) |
| Allocated bytes | 64,353,075 | 64,572,745 | 64,484,745 |

rs545837 (approved):
LGTM. Verified locally — all 406 tests pass, allocation counts are deterministic across runs and match the PR's claims exactly.

One minor follow-up opportunity (can be a separate PR): `consumed_words` can be `HashSet<&str>` instead of `HashSet<String>` since the inserted `member`/`receiver` values are already borrowed `&str` from `dot_chains`. Saves another ~4,000 allocations on the same benchmark. Left details in a comment.

PR #346, opened 2026-06-08
Use faster hashing in graph resolution

### LINKED ISSUE #324: Low-effort global wins for the graph build: custom allocator and a faster hasher

Iron-Ham:
Closes #324

## Summary
- Uses `rustc_hash` for graph-build and scope-resolution hot maps while keeping public `EntityGraph` map fields on standard `HashMap` aliases.
- Keeps graph edge ordering deterministic by sorting resolved edges and routing cached incremental graph reconstruction through the same sorted `EntityGraph::from_parts` path.
- Generalizes import-resolution/test helpers so public callers can keep using standard `HashMap`s while internal graph paths avoid the std hasher in hot loops.

## Metrics
Environment: macOS, `RAYON_NUM_THREADS=8`, release CLI, generated mixed JS/TS fixture from `benchmarks/large-js-fixture`, 12 entities/file, fanout 5, nested depth 3, body lines 40. Baseline is installed `sem 0.8.0`; this PR is `crates/target/release/sem`. Five runs per size, median shown; OS file cache was not purged.

| Workload | Phase | Baseline median | PR median | Delta |
|---|---:|---:|---:|---:|
| 1,000 files | cold graph | 1970.985 ms | 1941.471 ms | 1.50% faster |
| 1,000 files | warm cache | 113.114 ms | 112.063 ms | 0.93% faster |
| 1,000 files | one-file incremental | 301.043 ms | 273.684 ms | 9.09% faster |
| 5,000 files | cold graph | 10740.003 ms | 10358.670 ms | 3.55% faster |
| 5,000 files | warm cache | 509.326 ms | 502.839 ms | 1.27% faster |
| 5,000 files | one-file incremental | 1699.274 ms | 1536.676 ms | 9.57% faster |

Correctness check: release `sem graph --json --no-cache` output stayed byte-identical to the origin-main baseline on generated fixtures:

| Fixture | SHA-256 | Result |
|---|---|---|
| 250 files | `7875c88d1905d6c8baba41d401e947fbdb0c6d0ebc8499da14a1a0e60ec2a455` | match |
| 1,000 files | `c439d5618d2b6156df317825660cc0116034ece4acbbd7c8a7f4449bf00907eb` | match |

## Test Plan
- `cargo test -p sem-core parser::graph::tests::test_incremental_import_candidates_re_resolve_clean_barrel`
- `cargo test -p sem-cli --test graph_json`
- `cargo test -p sem-cli cache::tests::`
- `cargo test -p sem-mcp cache::tests::`
- `cargo test --workspace`
- `cargo build --release -p sem-cli`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to sem are documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- Graph resolution now uses faster hash collections in hot paths to reduce graph build overhead.
+- Scope resolution caches repeated reference lookups during graph builds to reduce redundant resolver work.
+- Graph builds avoid retaining import scan source text after import extraction, reducing peak memory use.
+
 ### Fixed
 
 - Fixed: `super::module::func()` calls were dropped from the entity graph, so `impact` and `context` under-reported the blast radius across modules. Multi-segment Rust path-prefixed calls (`super::`/`crate::`/`self::`) now resolve to the real entity.
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1391,6 +1391,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustc-hash"
+version = "2.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94300abf3f1ae2e2b8ffb7b58043de3d399c73fa6f4b73826402a5c457614dbe"
+
+[[package]]
 name = "rustix"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1520,6 +1526,7 @@ dependencies = [
  "git2",
  "rayon",
  "regex",
+ "rustc-hash",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/crates/sem-cli/src/cache.rs
+++ b/crates/sem-cli/src/cache.rs
@@ -1050,7 +1050,7 @@ mod tests {
     }
 
     fn graph_with_edges(entities: &[SemanticEntity], edges: Vec<EntityRef>) -> EntityGraph {
-        let entity_map = entities
+        let entity_map: HashMap<String, EntityInfo> = entities
             .iter()
             .map(|entity| {
                 (
--- a/crates/sem-core/Cargo.toml
+++ b/crates/sem-core/Cargo.toml
@@ -115,6 +115,7 @@ tree-sitter-haskell = { version = "0.23", optional = true }
 tree-sitter-elm = { version = "5.9", optional = true }
 tree-sitter-clojure-orchard = { version = "0.2.5", optional = true }
 tree-sitter-d = { version = "0.8.2", optional = true }
+rustc-hash = "2"
 
 [dev-dependencies]
 tempfile = "3.24.0"
--- a/crates/sem-core/src/parser/context.rs
+++ b/crates/sem-core/src/parser/context.rs
@@ -5,7 +5,7 @@
 use std::collections::{HashMap, HashSet, VecDeque};
 
 use crate::model::entity::SemanticEntity;
-use crate::parser::graph::EntityGraph;
+use crate::parser::graph::{EntityAdjacencyMap, EntityGraph};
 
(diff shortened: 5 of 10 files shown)


rs545837:
ImportsByFile is Fx-hashed now, and maps with a custom hasher don't have ::new(). The line is in a Windows-only code path, which is why it slipped past macOS/Linux CI and local testing. 

Rest of the PR looks great, I checked it out locally and can see better hashing

PR #357, opened 2026-06-11
Multi-path `sem entities` (closes #356) + verify install.sh checksum

### LINKED ISSUE #356: Support multiple path arguments where appropriate
Commands that take a path should allow multiple paths, because AI agents like to be efficient and pass many paths at once.

Example:
```
$ sem entities core/daemon/test/support/temp.ts core/review-sync/test/support.ts core/sprint-branch/test/support.ts


error: unexpected argument 'core/review-sync/test/support.ts' found

Usage: sem entities [OPTIONS] [PATH]

For more information, try '--help'.
```

rs545837:
Two related CLI/distribution fixes in one PR.

**1. Multiple path arguments for `sem entities` (closes #356)**
`sem entities a.ts b.ts c.ts` previously errored with "unexpected argument". It now accepts any number of paths, extracts entities from each, and combines results deduped by entity id (so an overlapping directory + a file inside it does not double-count). Single-path and default-`.` behavior is unchanged, and the cloud whole-repo fast-path still applies only to the single-path case. Thanks @aleclarson for the report.

**2. Verify release checksum in install.sh**
The `curl | sh` installer downloaded and extracted the release tarball without verifying it, even though `checksums.txt` is published alongside. `sem update` already verifies; this mirrors it in the installer. Hard-fails on mismatch, skips silently when no sha256 tool or matching entry exists.

Both verified locally (multi-path/back-compat/dedup/error cases; `sh -n` plus a real temp install showing "Verified checksum").

### DIFF
--- a/crates/sem-cli/src/commands/cloud.rs
+++ b/crates/sem-cli/src/commands/cloud.rs
@@ -1027,8 +1027,14 @@ pub fn try_cloud_context(opts: &ContextOptions) -> Option<()> {
 
 /// Try to run `sem entities` via cloud (whole-repo directory listing only).
 pub fn try_cloud_entities(opts: &EntitiesOptions) -> Option<()> {
-    // Only use cloud for directory (whole-repo) listing, not single files
-    let path_arg = opts.path.as_deref().filter(|p| !p.is_empty()).unwrap_or(".");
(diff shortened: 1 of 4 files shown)


aleclarson:
awesome, thank you!

PR #391, opened 2026-06-24
feat: gate self-update behind a default-on Cargo feature (#390)

### LINKED ISSUE #390: feat: gate self-update behind a Cargo feature flag
Hi,

First of all, @0-wiz-0 packaged `sem` for NetBSD and I've merged the package into the pkgsrc main branch yesterday. You can find the commit log here: https://mail-index.netbsd.org/pkgsrc-changes/2026/06/23/msg348685.html

Today, I've updated it to v0.14.1, https://mail-index.netbsd.org/pkgsrc-changes/2026/06/24/msg348801.html

But one thing is concerning me as a package maintainer.

Could the self-update feature `sem update` be gated behind an optional Cargo feature so that we downstream package maintainers can opt out at build time?

When `sem` is installed via a system package manager (pkgsrc, Homebrew, Debian/apt, Fedora/rpm, etc.), the package manager owns the binary's lifecycle. Self-updating in that context causes several problems.

This would also make it easier for `sem` to be adopted by other packaging systems. Thanks!

rs545837:
Implements #390 (requested by @0323pin, pkgsrc/NetBSD maintainer).

Package managers own the binary's lifecycle, so a self-replacing `sem update` breaks distro builds. This puts `sem update` + the background update-available check behind a **`self-update` feature, on by default**:
- **curl/npm installs**: unchanged (feature on).
- **distro builds**: `cargo build --no-default-features` → `sem update` prints "update through your package manager" instead of overwriting the binary, and the background check is a no-op.

The real update module (ureq download, checksum verify, binary replace) only compiles with the feature; no-op stubs keep call sites compiling when off. `ureq` stays a dependency regardless (cloud + telemetry use it).

Verified: both `cargo build -p sem-cli` and `--no-default-features` compile; OFF binary prints the package-manager message, ON binary still does the real version check.

Closes #390.

(Related: a packager may also want telemetry behind a build-time opt-out, can follow up if @0323pin wants it.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to sem are documented in this file.
 
 ### Added
 
+- `self-update` Cargo feature (on by default) gates the built-in `sem update` and the background update-available check. Distro and package-manager builds that own the binary's lifecycle can opt out with `cargo build --no-default-features`; `sem update` then prints a "update through your package manager" message instead of replacing the binary. Thanks @0323pin (pkgsrc/NetBSD) for the request (#390).
+
+### Added
+
 - `sem context --hops N` bounds the related entities to N graph hops from the target (instead of filling to the token budget), so you can ask for "the entity and just its immediate neighborhood." The `sem_context` MCP tool gains the same `hops` parameter. 0 (the default) keeps the existing unbounded, budget-driven behavior.
 
 ### Changed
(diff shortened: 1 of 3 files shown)


0323pin:
@rs545837 

> Related: a packager may also want telemetry behind a build-time opt-out, can follow up if @0323pin wants it.

I took the patch from FreeBSD that reverts the telemetry logic and applied that [on our build](https://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/devel/sem/patches/patch-src_telemetry.rs?rev=1.1), so for now that's fine. Thanks!

PR #394, opened 2026-06-28
feat: Lua support (#393) + byte offsets on SemanticEntity

### LINKED ISSUE #393: Support for Lua
Great tool! Using it heavily and would love if support for [lua](https://github.com/tree-sitter-grammars/tree-sitter-lua) could be added for better script writing, esp. for neovim. There's also a [luadoc](https://github.com/tree-sitter-grammars/tree-sitter-luadoc) grammar as well.

rs545837:
Two user-requested sem-core additions.

## Lua support (closes #393)
Adds the Lua grammar via `tree-sitter-lua` (tree-sitter-grammars/tree-sitter-lua, the de-facto standard maintained grammar @mmgeorge linked), gated behind a `lang-lua` feature in `grammar-all`. One node type, `function_declaration`, covers global, local, table (`t.f`) and method (`t:f`) forms. Verified with `sem entities` / `sem diff` on real Lua.

## Byte offsets on SemanticEntity
Adds optional `start_byte`/`end_byte`, populated from the tree-sitter node during code extraction, so a consumer can slice the exact original bytes of an entity out of a file given only `file_path` + the span (no re-parsing). `None` for parsers that don't expose byte spans. Persisted through the sem-mcp/sem-cli SQLite entity cache (two nullable columns, schema v6→v7) and surfaced in `sem entities --json`. Requested by a sem-core user (Thomas J).

Tests: full suites pass (sem-core 404, sem-mcp 55); regression tests for both. Note: public-API addition to `SemanticEntity` → minor version bump at release.

### DIFF
--- a/.claude/hooks/sem-first.sh
+++ b/.claude/hooks/sem-first.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+"""PreToolUse hook: make sem the prevalent tool for code-symbol search.
+
+When the agent runs the Grep tool with a bare code identifier (a function /
+type / variable name), block it and steer to the sem MCP tools, which answer
+structural questions grep cannot (cross-file callers/callees, blast radius)
+without false positives.
+
+Deliberately NOT a hard wall:
+  - Only Grep-tool calls whose pattern is a bare identifier are redirected.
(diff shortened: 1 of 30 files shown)


mmgeorge:
Thanks for adding this!

PR #398, opened 2026-06-30
perf(cache): parallelize per-file freshness scan (#351)

### LINKED ISSUE #351: Parallelize cache freshness checks for touched-file cache hits

## Summary

Follow-up from #344: after batched fingerprint refreshes, touched-file cache hits still spend most of their remaining time in the sequential freshness scan.

The review benchmark isolated `cache_topology_load` as:

| Scenario | `cache_topology_load` |
| --- | ---: |
| Pure cache hit, mtimes match | ~6 ms |
| All 5K files touched | ~48 ms |

That leaves roughly 42 ms in sequential filesystem/hash work: one `file_freshness()` check per source file, doing `stat()` + `read()` + xxhash when mtimes changed. The checks are independent, so a parallel scan over the file list could reduce the touched-file cache-hit path on multi-core machines.

Source follow-up: https://github.com/Ataraxy-Labs/sem/pull/344#issuecomment-4668480041

## Duplicate Check

Checked open PRs and issues before filing. Existing open performance work is focused on graph-build hashing/allocation, import-resolution allocation, source retention, and scope-resolution costs; none covers parallel cache freshness checks or `file_freshness()` parallelism.

Targeted open-tracker searches for `file_freshness`, `cache_topology_load`, `freshness`, and `rayon cache` also returned no matches.

## Current Shape

The relevant cache paths are the freshness loops in the CLI and MCP caches, including:

- `sem-cli` full/topology cache freshness via `has_fresh_cache`
- `sem-cli` partial cache freshness via `load_partial`
- `sem-mcp` topology cache freshness via `load_graph_topology`
- `sem-mcp` partial cache freshness via `load_partial`

The batch refresh write should remain best-effort and should still happen after freshness results are collected.

## Candidate Direction

Evaluate using `rayon::par_iter()` or a bounded worker pool over the source file list for filesystem/hash freshness checks. Keep SQLite access out of the parallel section: read cached metadata before the scan, collect stale-file decisions and fingerprint refreshes from workers, then apply refresh writes serially or through the existing batched best-effort helper.

`sem-core` already has optional `rayon`, but the cache loops live in `sem-cli` and `sem-mcp`, so part of the work is deciding where the dependency/feature boundary should sit.

## Acceptance Criteria

- Freshness checks for large touched-file cache hits run in parallel where it is safe to do so.
- Existing cache semantics are preserved: manifest entries are skipped, missing/stale files still miss, and failed auxiliary fingerprint refresh writes do not invalidate otherwise valid cache hits.
- CLI and MCP cache paths have regression coverage for fresh, stale, and `FreshWithUpdatedFingerprint` cases.
- Benchmark the same shape as #344: prime a cache, touch many files without changing content, and compare `cache_topology_load` / total command time before and after with a fixed thread count.
- Full workspace tests pass.

rs545837:
Closes #351. Follow-up from #344 (batched fingerprint refreshes).

After #344 landed, touched-file cache hits still spent most of their remaining time in a sequential freshness scan: one `file_freshness()` per source file doing `stat()` + `read()` + xxhash. The #344 benchmark isolated this as roughly 42ms of sequential filesystem/hash work on a 5K-file touched scenario, the dominant remaining cost on that path.

The checks are independent, so this parallelizes them with rayon while keeping SQLite strictly out of the parallel section. `cached_files_are_fresh` now runs in three phases:

1. **Serial (SQLite):** read each file's cached fingerprint. The rusqlite connection isn't shareable across threads, so all DB reads happen here. A file missing from the cache bails immediately, same as before.
2. **Parallel (rayon, no SQLite):** `file_freshness` is a pure `stat`+`read`+hash, so it runs across the pool via `into_par_iter`.
3. **Serial:** one stale file fails the whole cache; otherwise refreshed fingerprints are persisted best-effort (the write stays serial, as before).

Cache-hit validity is unchanged. Adds `rayon` to sem-cli.

### Testing
- Full sem-cli suite passes (99 unit + cache tests, plus the cache-freshness tests).
- The 2 `cwd_resolution` failures are pre-existing on `main` (an unrelated `diff --patch` pathspec test); confirmed by running them on a clean tree. Not introduced here.

### Scope
This addresses #351 only. The other open perf issues, #320 (intern entity IDs) and #322 (share file buffers vs per-entity content), are larger core refactors with a big public-API blast radius that need ~27GB-scale RSS profiling to validate, so they are intentionally out of this PR.

Thanks @Iron-Ham for the issue and the benchmark.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All notable changes to sem are documented in this file.
 - `sem xref` lists cross-repo dependencies across your indexed repos: entities in one repo that depend on entities in another. A single-repo local graph can't see this, so it's a cloud feature (requires `sem login`) and is gated to the team/enterprise tier. Adds `cross_deps()` to the shared cloud client.
 - `sem diff` now prints a one-line hint, when run interactively and logged out, that `sem login` reveals what your changes break across repos (a cross-repo question a local single-repo diff can't answer). It is heavily throttled (at most once a week), shown only on a terminal with real entity changes, and stays completely silent in CI, pipes, `--json`/non-terminal output, and for logged-in users.
 
+### Performance
+
+- Cache freshness checks now run the per-file `stat` + content-hash scan in parallel (rayon) instead of sequentially (#351). On touched-file cache hits over large repos, the freshness scan was the dominant remaining cost (~42ms of sequential filesystem/hash work on a 5K-file touched scenario); it now scales across cores. SQLite reads stay serial (the connection isn't shared across threads) and fingerprint-refresh writes remain serial and best-effort — only the pure filesystem+hash work is parallelized, so cache-hit validity is unchanged.
+
 ### Documentation
 
 - README: documented the optional cloud acceleration flow (`sem login` serves `impact`/`context`/`entities` from a warm pre-built graph for large repos; local is unchanged and `SEM_LOCAL=1` forces local), and added Lua to the supported-languages table.
--- a/crates/sem-cli/Cargo.toml
+++ b/crates/sem-cli/Cargo.toml
@@ -27,6 +27,7 @@ ignore = "0.4"
 openssl-sys = { version = "0.9", optional = true }
 serde_json = "1"
 rusqlite = { version = "0.32", features = ["bundled"] }
+rayon = "1.10"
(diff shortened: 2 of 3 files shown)


Iron-Ham (approved):
mvp

PR #405, opened 2026-07-01
fix: support relative worktree repositories

### LINKED ISSUE #341: Support for Git Worktrees
Really cool tool, thanks for sharing!

Support for Git Worktrees would be useful. Currently results in:

```
error: git error: unsupported extension name extensions.relativeworktrees; class=Repository (6)
```

zerone0x:
## Summary
Fixes #341.

Allows sem-core to open repositories whose config declares `extensions.relativeworktrees`, which libgit2 otherwise rejects as an unknown extension while discovering worktrees.

## Changes
- Registers `relativeworktrees` as a supported libgit2 extension before repository discovery.
- Adds a regression test covering repository open with `extensions.relativeworktrees` enabled.

## Test Plan
- `rustfmt sem-core/src/git/bridge.rs --check`
- `cargo test -p sem-core open_allows_relative_worktrees_extension --features git --no-default-features` *(blocked in this environment: missing system C linker `cc`)*


rs545837:
Merged, thank you @zerone0x! Great pickup on #341.

One note: I pushed a small fixup to your branch before merging. There was a Result type mismatch that did not compile (.as_ref() left a Result<&(), GitError>, needed a .map(|_| ()) to get back to unit), plus the comment said Git 2.41 but relativeWorktrees landed in Git 2.48. Since your environment could not build, we ran the verification here: the regression test genuinely fails without the fix and passes with it, and the full sem-core suite is green (405 tests).

This ships in the next release.

PR #462, opened 2026-07-05
hook: inject an orient briefing for plain-English tasks

rs545837:
The `sem hook prompt-submit` UserPromptSubmit hook resolved only identifier-shaped tokens from the prompt by **exact name** — so it fired only when the user literally typed a `function_name`, and stayed silent on ordinary tasks like *"fix the bug where empty blueprint names don't raise an error"*, which is exactly when injecting the code would help most.

**Fix:** when no entity is named, fall back to an `orient --pack` briefing (tight, 3 entities) over the whole prompt, so the agent starts with the relevant code in context and skips the search. This is the 0-turn understanding path — measured at 9.4s vs 13.2s (−29%) when the code is pre-injected, vs the agent *calling* a tool (a turn) which was net slower.

Verified locally:
- Coding/navigation prompts inject a briefing (`fix the bug where empty blueprint names…`, `where is the code that builds the graph`).
- Chat/meta injects nothing (`explain the philosophy…`, `are we faster`, `…the fundraise` — whole-word match so 'fundraise' doesn't trip 'raise').
- Fast: ~0.05s on a cached repo, hard-capped at 2.5s so a cold/huge repo can't stall a prompt.

This is the 'last mile' that makes auto-injection actually fire on real tasks, built on the ranking fixes in #460/#461.


rs545837:
Superseded by the orient removal (#463). That PR added the fuzzy hook fallback; we're removing all fuzzy retrieval instead.

PR #492, opened 2026-09-22
fix: Dart dependency graphs and cross-language caller isolation

### LINKED ISSUE #491: Dart: entities parse but the dependency graph has zero edges (impact/callers/refs silently empty)
## Summary

Dart entity extraction works, but Dart produces **zero dependency-graph edges**. `sem entities` and `sem diff`/`blame`/`log` are correct; `sem graph`, `impact`, `callers` and `refs` all behave as though the file had no references in it. Because `sem impact` answers `✓ No other entities are affected by changes to this entity`, the failure is silent and reads as a positive result — an agent or a human takes it as clearance to refactor.

## Repro

One repo, the same two-file shape in Dart and TypeScript:

```
dart/greeter.dart   class Greeter { String hello(String name) => 'hello $name'; }
dart/app.dart       import 'greeter.dart';  void run() { final g = Greeter(); print(g.hello('world')); }
ts/greeter.ts       export class Greeter { hello(name: string): string { return `hello ${name}`; } }
ts/app.ts           import { Greeter } from './greeter';  export function run(): void { const g = new Greeter(); console.log(g.hello('world')); }
```

`sem entities .` is symmetric — all four files parse, same entity kinds:

```
  dart/app.dart        function run (L3:6)
  dart/greeter.dart    class Greeter (L1:3) / method hello (L2:2)
  ts/app.ts            function run (L3:6)
  ts/greeter.ts        class Greeter (L1:3) / method hello (L2:2)
```

`sem graph` is not:

```
⊕ 6 entities, 2 edges
```

Both edges come from the TypeScript pair. Dart contributes none:

```
$ sem impact --entity-id 'dart/greeter.dart::class::Greeter'
⊕ class Greeter (dart/greeter.dart:1–3)
  ✓ No other entities are affected by changes to this entity.

$ sem impact --entity-id 'ts/greeter.ts::class::Greeter'
⊕ class Greeter (ts/greeter.ts:1–3)
  ← depended on by:
    ← function run (ts/app.ts)
```

In a Dart-only repo it's unambiguous: `sem graph` → `3 entities, 0 edges`, and `callers`/`refs` both return `(callers: none)` / `(refs: none)`.

## Likely cause

`crates/sem-core/src/parser/plugins/code/languages.rs` registers Dart, but the resolution wiring is empty or points at the wrong node:

`DART_CONFIG` (L959) — `call_entity_identifiers: &[]`, `scope_boundary_types: &[]`.

`DART_SCOPE_CONFIG` (L2443):

| field | Dart | TS | Kotlin |
|---|---|---|---|
| `call_nodes` | `["function_expression_body"]` | `["call_expression"]` | `["call_expression"]` |
| `assignment_rules` | `&[]` | populated | populated |
| `member_access` | `&[]` | populated | populated |
| `new_expr_nodes` | `&[]` | `["new_expression"]` | `&[]` |

`function_expression_body` is the `=> expr` arrow-body node, not a call site, so no call is ever matched. With `assignment_rules` empty, `final g = Greeter()` never binds `g` to `Greeter`; with `member_access` empty, `g.hello(...)` is never resolved as a member call.

**Kotlin looks like the right template** — it also has no `new` keyword and resolves construction through `call_nodes: ["call_expression"]` plus `assignment_rules` and `member_access`. Dart should need the same three fields plus a Dart-appropriate `call_nodes`.

## Second, related defect: `callers` leaks across files and languages

In the mixed repo above:

```
$ sem callers 'hello' --file dart/greeter.dart
method hello dart/greeter.dart:2
  function run ts/app.ts:3
```

The target resolves correctly to the Dart method, but the reported caller is in a **different file and a different language**. So the Dart gap isn't only a false negative — the fallback produces confident false positives too. Happy to split this into its own issue if you'd prefer.

## Environment

sem 0.25.0 (latest), macOS arm64 (Darwin 24.6.0), installed via Homebrew `sem-cli`.

rs545837:
## Summary
Fixes #491.

- Resolve real Dart call expressions, declaration scopes, constructor-bound locals and typed parameters, including new/const construction.
- Filter typed/static member candidates by language family; prefer imported/local class owners. Preserve JS/TS interoperability.
- Invalidate persisted graph/query caches so upgrades do not retain the old edges.
- Add eight core regression tests and cold/warm CLI callers/refs coverage. Replace the old test that expected zero Dart call edges.

## Validation
- Local core/CLI unit and integration suites: 1,115 passing tests across 44 suites.
- Final Dart fixture rerun: all eight tests pass.
- Core doc-test target passes (zero doctests).
- Tests ran with GIT_CONFIG_GLOBAL=/dev/null and GIT_CONFIG_NOSYSTEM=1 to prevent the locally configured external Git diff wrapper from corrupting patch-input fixtures. No machine settings changed.
- CI is pending; merge is gated on the latest head and passing checks.

## Boundaries
This fixes the reported extraction and mixed-language collision bugs, not complete Dart semantic resolution. Dart import aliases, package URIs, dynamic dispatch and incremental resolution reuse require separate work. Dart remains conservatively ineligible for incremental edge reuse.


rs545837:
Temporarily reopening to refresh a stale PR head: the branch points to 1a9686ad4c64de49e3b9cd809fdea302bdc67e0f, while the PR ref is still f5da7edb8f999921f85af5fb0d284cd7db37d7cd. The latest commits include the required changelog entry and a fixture correction. No merge until checks cover the current head.

rs545837:
Refreshing the stale PR head after the final explicit-this regression fix. Final local validation: 1,115 core/CLI tests pass, including eight focused core regressions and cold/warm CLI coverage. The branch is now 2c427f2; CI must evaluate that head.

